PR #543, opened 2021-09-08
436 Updates to use Sys.timezone()

aaehmann:
Updated `as_iso_dtm` to use the `Sys.timezone()`

I needed to udpate the test scripts as well to assign the time zone to the system timezone.  

I think we need to revisit these date functions.  I am thinking perhaps for all dates that are passed in to be used in manipulations (exclude where they are sorts and things like that) we have a function or rule to `lubridate::force_tz(date, tz = "UTC")`.  We are assigning a timezone in `impute_dtc` as "UTC" and that may be the start of everything getting a bit our of sync, those dates are interacting with dates passed in.  We may want to revisit this again.

I checked in our WARP again and with the update to the test scripts, it seems to be working.

### DIFF
--- a/R/iso_dtm.R
+++ b/R/iso_dtm.R
@@ -1,4 +1,4 @@
-as_iso_dtm <- function(x, time_zone = "UTC") {
+as_iso_dtm <- function(x, time_zone = Sys.timezone()) {
   if (is.character(x)) {
     dtm <- ymd_hms(x, tz = time_zone)
   } else {
--- a/tests/testthat/test-derive_vars_dtm.R
+++ b/tests/testthat/test-derive_vars_dtm.R
@@ -14,10 +14,10 @@ input <- tibble::tribble(
 test_that("default: no date imputation, time part set o 00:00:00, add DTF, TMF", {
   expected_output <- tibble::tribble(
     ~XXSTDTC, ~ASTDTM, ~ASTTMF,
-    "2019-07-18T15:25:40", ymd_hms("2019-07-18T15:25:40"),  NA_character_,
-    "2019-07-18T15:25", ymd_hms("2019-07-18T15:25:00"), "S",
-    "2019-07-18T15", ymd_hms("2019-07-18T15:00:00"), "M",
-    "2019-07-18", ymd_hms("2019-07-18T00:00:00"), "H",
+    "2019-07-18T15:25:40", ymd_hms("2019-07-18T15:25:40", tz=Sys.timezone()),  NA_character_,
+    "2019-07-18T15:25", ymd_hms("2019-07-18T15:25:00", tz=Sys.timezone()), "S",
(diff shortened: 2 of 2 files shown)


kabis-ops:
Hi @aehmann-gsk, 
As discussed yesterday, I had the same kind of issue where the derive_vars_dtm (as_iso_dtm) was adding 2 hours to my original DTC. Thank you for sharing the workaround. I have tried it but it does not always work: I had to also set the time zone when the variable is not character and it works.
```
as_iso_dtm2 <- function(x, time_zone = Sys.timezone()) {
  dtm <- ymd_hms(x, tz = time_zone)

  class(dtm) <- union("iso_dtm", class(dtm))
  dtm
}
```

You can test it with:
```
tibble::tribble(
  ~XXSTDTC, ~ASTDTM, ~ASTTMF,
  "2019-07-18T15:25:40", ymd_hms("2019-07-18T15:25:40"),  NA_character_,
  "2019-07-18T15:25", ymd_hms("2019-07-18T15:25:00"), "S",
  "2019-07-18T15", ymd_hms("2019-07-18T15:00:00"), "M",
  "2019-07-18", ymd_hms("2019-07-18T00:00:00"), "H",
) %>%
  mutate(ASTDTM2a = ASTDTM %>% as_iso_dtm(),
         ASTDTM2b = ASTDTM %>% as_iso_dtm2(),
         ASTDTM3 = impute_dtc(XXSTDTC,time_imputation = "00:00:00") %>% as_iso_dtm2()
  )
```
image

ASTDTM2a (using the proposed updated function) has still +2 (for me) hours
ASTDTM2b (using the corrected `as_iso_dtm2())` function is correct. 
ASDTM3 was for testing for character date input.

Please let me know if that also works for you?

kabis-ops:
Hi again,
Just to add that this happens only when i work outside of the admiral project. When i run this in the admiral project the update you  have made works fine and i dont need the as_iso_dtm2() function 🤔.

aehmann-gsk:
Hi @kabis-ops -
Thank you!  I will test this today.  I had different results when I ran in our production GSK production environment vs. my local laptop within the Admiral project.  I had to change the test script to assign the timezone when passing a date to get around this issue when running locally.  It looks like your solution based on your input/output above may correct this issue. 👍 

Overall, this issue makes me nervous as this function if forcing in the user's timezone but in impute_dtc it is forcing in UTC as a timezone in 2 lines of code.  In general, what would happen if you and I are working on the same study, I write out a permanent dataframe and then you read that dataframe in and use.  Do the dates look different for you and I?  I will try to test this scenario out as well.

koegerr:
@aehmann-gsk would it help if I connect you with one or two R experts at Roche for this? Maybe they could help.

aehmann-gsk:
@koegerr Yes - I think that will help.  I also need to connect with GSK experts.  thanks!

PR #555, opened 2021-09-15
1st draft of DTM to TM

teckla-gsk:
a few comments, I have included a hms flag to allower users to toggle between HH:MM or HH:MM:SS. Also, TM-vars are classed character since posixct/lt are for DTM. I could use the hms package but cannot seem to find an option to only display HH:MM the hms only allows for HH:MM:SS. Of course, open to any alternative solutions


R/derive_vars_dtm_to_tm.R
```diff
@@ -0,0 +1,64 @@
...
+#'  ~STUDYID, ~USUBJID, ~TRTSDTM,               ~ASTDTM,               ~AENDTM,
+#'  "TEST01", "PAT01",  "2012-02-25 23:41:10", "2012-02-28 19:03:00", "2013-02-25 23:32:16",
+#'  "TEST01", "PAT01",  "",                    "2012-02-28 19:00:00", "",
+#'  "TEST01", "PAT01",  "2017-02-25 23:00:02", "2013-02-25 19:00:15", "2014-02-25 19:00:56",
+#'  "TEST01", "PAT01",  "2017-02-25 16:00:00", "2017-02-25 14:25:00", "2017-03-25 23:00:00",
+#'  "TEST01", "PAT01",  "2017-02-25 16:05:17", "2017-02-25 14:20:00", "2018-04-29 14:06:45",
+#'  ) %>% mutate(
+#'  TRTSDTM = lubridate::as_datetime(TRTSDTM),
+#'  ASTDTM = lubridate::as_datetime(ASTDTM),
+#'  AENDTM = lubridate::as_datetime(AENDTM)
```
thomas-neitmann, line 37: No need to use the `lubridate::` prefix here as you've already loaded the package above.

R/derive_vars_dtm_to_tm.R
```diff
@@ -0,0 +1,64 @@
+#' Derive time variables from datetime variables
```
thomas-neitmann, line 1: Please use Title Case here.

R/derive_vars_dtm_to_tm.R
```diff
@@ -0,0 +1,64 @@
...
+#' derive_vars_dtm_to_tm(adcm,
+#'                       vars(TRTSDTM, ASTDTM, AENDTM))
+#'
+#' derive_vars_dtm_to_tm(adcm,
+#'                      vars(TRTSDTM, ASTDTM, AENDTM),
+#'                      hms_flag="Y")
+
+derive_vars_dtm_to_tm <- function(dataset,
+                                  source_vars,
+                                  hms_flag=NULL) {
```
thomas-neitmann, line 49: If we want to keep `hms_flag` then I'd suggest to use `TRUE` and `FALSE` rather than `"Y"` and `NULL`. However, I'm not quite sure about this functionality. I would expect to get the "whole" time component from the datetime rather than only hour and minute. This seems to be more about display (kind of like a SAS format) rather than the actual data. I'd suggest to make use of the `hms` package. Any thought @aehmann-gsk @bundfussr?

thomas-neitmann (changes requested):

R/derive_vars_dtm_to_tm.R
bundfussr: According to ADaMIG `*TM` variables must be numeric. I.e., we should update the function such that the new variables are numeric and then we no longer need the `hms_flag`.

bundfussr (changes requested):

R/derive_vars_dtm_to_tm.R
teckla-gsk: This is a bit of a challenge since for numeric times, they are being displayed as number of seconds since 1970-01-01. I can implement this using the hms package resulting in class "hms" and displaying as HH:MM:SS or the initial function resulting in class "character" with option to display as HH:MM or HH:MM:SS. Otherwise, I do not know that R can handle times akin to SAS where it is numeric but displays as HH:MM. Open to exploring other packages/suggestions

bundfussr: The `hms` package looks like what we need. @thomas-neitmann , @aehmann-gsk , do you agree?

Do you now if the `hms` class becomes a time variable when exported as a SAS dataset?

teckla-gsk: Updated to use hms. So far, I have not successfully read in the time vars from the xpt. They do not compute, but will explore more

kabis-ops: Hi @teckla-gsk , @bundfussr, @esimms999-gsk 

I have make a try with the `haven` package and there seems to be no issue reading the date nor time in XPT/ SAS
```
library(haven) 
library(diffdf)

input <- tibble::tribble(
  ~STUDYID, ~USUBJID, ~TRTSDTM, ~ADTM,               
  "TEST01", "PAT01",  "2012-02-25 23:41:10", "2012-02-28 19:03:00") %>% 
  mutate(TRTSDTM = as_datetime(TRTSDTM), ADTM = as_datetime(ADTM))

actual_output <- derive_vars_dtm_to_tm(input, vars(TRTSDTM, ADTM))

#Convert to XPT
write_xpt(actual_output, "test.xpt",version = 5)           

#Compare data before xpt with data read from xpt
actual_output_fromxpt<-read_xpt("test.xpt")     

diffdf(actual_output, actual_output_fromxpt)
```
The only diff between the 2 dataframes is the `format.sas` attributes, and I guess this will be dealt with once we tackle the metadata stuff.

```
There are columns in BASE and COMPARE with differing attributes !!
All rows are shown in table below

  ================================================
   VARIABLE  ATTR_NAME   VALUES.BASE  VALUES.COMP 
  ------------------------------------------------
     ADTM    format.sas     NULL       DATETIME   
     ATM     format.sas     NULL         TIME     
   TRTSDTM   format.sas     NULL       DATETIME   
    TRTSTM   format.sas     NULL         TIME     
  ------------------------------------------------
```

I was also been able to read the XPT file in SAS with no problem
```
libname xptfile xport 'Path\test.xpt' access=readonly;
proc copy inlib=xptfile outlib=work;
run;
```
image

So i guess we are good :)

Maybe 1 comment regarding the function itself: i have first tried with 1 DTM variable as input, but when only 1 DTM is specified the new variable is misspelled as **'new'**
image

kabis-ops (commented):
Hi, Please see [here](https://github.com/Roche-GSK/admiral/pull/555#discussion_r720049662)

R/derive_vars_dtm_to_tm.R
teckla-gsk: @kabis-ops thanks for testing it and the comment. I will update it based on your comment :)

thomas-neitmann:
@teckla-gsk Please declare the `hms` package as a dependecy of `admiral` by listing it under the `Imports` section in the `DESCRIPTION` file. That should get rid of the errors when running `R CMD check`.

teckla-gsk:
@kabis-ops please see update to handle when a single var is parsed, I suppose I also have to update derive_vars_dtm_to_dt

teckla-gsk:
@thomas-neitmann  I added hms to the description file but the RMD is still failing. Any ideas as to what I am missing here?

DESCRIPTION
```diff
@@ -32,6 +32,7 @@ Imports:
     purrr,
```
teckla-gsk, line 32: @thomas-neitmann  still receiving rmd fail despite adding hms to decription file here

bundfussr (changes requested):
Could you update NEWS.md?

PR #577, opened 2021-10-04
Add section on handling missing values

### LINKED ISSUE #531: Add `NA` Section to "Get Started" Vignette
A section regarding `NA` should be added to the "Get Started" vignette. It should state:
- All admiral functions expect that `NA` instead of `""` is used.
- admiral provides the `convert_blanks_to_na()` function to ensure that the input datasets fulfill the requirement.
- some examples of condition where you should take care, e.g., `DTYPE != "AVERAGE" | is.na(DTYPE)"` must be used instead of `DTYPE != "AVERAGE"`.

thomas-neitmann:
Close #531

### DIFF
--- a/vignettes/admiral.Rmd
+++ b/vignettes/admiral.Rmd
@@ -57,7 +57,7 @@ the variable name, e.g., `derive_var_trtdurd()` derives the `TRTDURD` variable.
 Functions which derive a dedicated parameter start with `derive_param_` followed by
 the parameter name, e.g., `derive_param_os()` derives the `OS` parameter.
 
-## Input and Output
+# Input and Output
 
 It is expected that the input dataset is not grouped. Otherwise an error is
 issued.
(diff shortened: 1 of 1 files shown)


thomas-neitmann:
@bms63 @yangc71 Would be great if you could review this PR. If anything in the section I've added is unclear please let me know. That means I did a bad job 😉

yangc71 (commented):
The Handling of Missing Values section is clear to me :). Should we also include reminders for functions such as min() max() that will return NA when NA is present if user doesn't specify `na.rm = T`?

thomas-neitmann:
> Should we also include reminders for functions such as min() max() that will return NA when NA is present if user doesn't specify na.rm = T?

That's a great suggestions!

thomas-neitmann:
@yangc71 I've incoprporated your suggestion. Feel free to approve this PR if it looks good to you 🙂

yangc71 (approved):

PR #580, opened 2021-10-04
579 update derive vars dtm to dt

teckla-gsk:
updated to handle when only one is being parsed


teckla-gsk:
sorry accidently created the pull request to master

PR #581, opened 2021-10-04
updated to handle when only one var is parsed

teckla-gsk:
updated to display correct var name when only one dtm is parsed

### DIFF
--- a/R/derive_vars_dtm_to_dt.R
+++ b/R/derive_vars_dtm_to_dt.R
@@ -35,11 +35,28 @@
 #'   )
 #'
 #' derive_vars_dtm_to_dt(adcm, vars(TRTSDTM, ASTDTM, AENDTM))
+
 derive_vars_dtm_to_dt <- function(dataset, source_vars) {
   assert_vars(source_vars)
   assert_data_frame(dataset, required_vars = source_vars)
 
-  dataset %>%
-    mutate_at(source_vars, .funs = list(new = lubridate::date)) %>%
-    rename_at(vars(ends_with("new")), .funs = ~str_remove(., "M_new"))
+  #Warn if --TM variables already exist
+  dtm_vars <- quo_c(source_vars)
+  dtm_vars2 <- vars2chr(dtm_vars)
+  n_vars <- length(dtm_vars)
+
+  for (i in n_vars) {
+    dt_vars <- str_replace(dtm_vars2[[i]], "DTM", "DT")
+    warn_if_vars_exist(dataset, dt_vars)
+  }
+
+  if (n_vars > 1) {
+    dataset %>%
+      mutate_at(source_vars, .funs = list(new = lubridate::date)) %>%
+      rename_at(vars(ends_with("new")), .funs = ~str_replace(., "DTM_new", "DT"))
+  }
+  else{
+    dataset %>%
+      mutate(!!sym(dt_vars) := lubridate::date(!!sym(dtm_vars2)))
+  }
 }
--- a/tests/testthat/test-derive_vars_dtm_to_dt.R
+++ b/tests/testthat/test-derive_vars_dtm_to_dt.R
@@ -1,5 +1,6 @@
 context("test-derive_vars_dt")
 
+#test 1 - multiple vars
 test_that("Convert a complete -- DTM into a date object", {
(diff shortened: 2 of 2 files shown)


R/derive_vars_dtm_to_dt.R
```diff
@@ -35,11 +35,28 @@
 #'   )
```
kabis-ops, line 35: Sorry i cannot comment on the grey code so, adding them here:
- (line1) The title should be Title Case.
- (line 3) Should we update as :This function creates date(s) as output from datetime variable(s)
- (line 11-13) : Should we update as : A data frame containing the input dataset with the corresponding date (`--DT`) variable(s) of all datetime variables (`--DTM`) specified in `source_vars`.
- line 24-30: i would suggest to remove `ASTTMF`

tests/testthat/test-derive_vars_dtm_to_dt.R
```diff
@@ -40,3 +41,43 @@ test_that("Convert a complete -- DTM into a date object", {
     keys = c("STUDYID", "USUBJID", "TRTSDTM", "ASTDTM", "AENDTM")
   )
 })
```
kabis-ops, line 43: same here 
- line 19: expected_output; I would remove TRTSDT, ASTDT, AENDT from the tribble and just create them in the mutate step from the DTM variable. I would also probably make 1 subject per row otherwise it looks strange to have multiple trt start for the same pt 
- line41: if you make 1 subject per row then the keys would need to be updated

tests/testthat/test-derive_vars_dtm_to_dt.R
```diff
@@ -1,5 +1,6 @@
 context("test-derive_vars_dt")
 
+#test 1 - multiple vars
```
kabis-ops, line 3: Should you remove the comment and update the description in testthat?

tests/testthat/test-derive_vars_dtm_to_dt.R
```diff
@@ -1,5 +1,6 @@
...
 test_that("Convert a complete -- DTM into a date object", {
```
kabis-ops, line 4: Suggest: "Convert a list of complete -- DTM into dates object"?

tests/testthat/test-derive_vars_dtm_to_dt.R
```diff
@@ -40,3 +41,43 @@ test_that("Convert a complete -- DTM into a date object", {
     keys = c("STUDYID", "USUBJID", "TRTSDTM", "ASTDTM", "AENDTM")
   )
 })
+
+#test 2 - single var
+test_that("Convert a complete -- DTM into a date object", {
```
kabis-ops, line 46: same as above:
Suggest: "Convert a single -- DTM into a date object"?

tests/testthat/test-derive_vars_dtm_to_dt.R
```diff
@@ -40,3 +41,43 @@ test_that("Convert a complete -- DTM into a date object", {
...
+    "TEST01", "PAT01", "2017-02-25 16:00:00", "2017-02-25 14:00:00", "2017-03-25 23:00:00",
+    "TEST01", "PAT01", "2017-02-25 16:00:00", "2017-02-25 14:00:00", "2018-04-29 14:00:00",
+  ) %>% mutate(
+    TRTSDTM = lubridate::as_datetime(TRTSDTM),
+    ASTDTM = lubridate::as_datetime(ASTDTM),
+    AENDTM = lubridate::as_datetime(AENDTM)
+  )
+
+  # nolint start
+  expected_output <- tibble::tribble(
```
kabis-ops, line 61: same comment as above

kabis-ops (commented):
HI @teckla-gsk , thanks this is now working fine with one variable too 👍 
I have only made cosmetics comments.

thomas-neitmann:
@teckla-gsk Is this realted to your other PR #555? If so, idelly you would have made the changes directly there (you can always just push new commits to an existing PR). Should we close #555 in favor of this PR?

teckla-gsk:
@thomas-neitmann this was to update the dates PR #555 is for the times. Is this ready to be merged?

thomas-neitmann (approved):

PR #582, opened 2021-10-05
Update @seealso to use [function()] format.

yangc71:
Hope I did this right. There are also instances other than @seealso statements where e.g. backticks are added inside [] for function reference, but I did not update those because they resolve to links correctly. Should we add a section under the programming strategy vignette to define a consistent format? Thanks!

### DIFF
--- a/R/call_derivation.R
+++ b/R/call_derivation.R
@@ -21,7 +21,7 @@
 #'
 #' @export
 #'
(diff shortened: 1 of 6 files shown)


thomas-neitmann (approved):
Looks about right to me 😉 

Thanks a lot!

PR #584, opened 2021-10-07
519 Scripts updated and VADs added

millerg23:
Some comments/questions:
ADVS/ADEG: We use AVALU, however this is not a CDISC derivation, should it be removed?
ADVS/ADEG: I moved adding of PARAM/PARAMN to near end of program to avoid carrying through, and re-calculating for derived PARAMCDs (MAP, BMI, BSA).
ADAE/ADCM: Should ADURU be populated when ADURN is NA? Maybe Roche process, but we only set ADURU for non-missing ADURN.
(Still to update Vignettes).


inst/templates/ad_adcm.R
```diff
@@ -16,16 +16,20 @@ library(lubridate)
 data("cm")
 data("adsl")
 data("ex")
-data("adsl")
```
thomas-neitmann, line 19: This should not be removed as you use `adsl` below.

inst/templates/ad_adcm.R
```diff
@@ -16,16 +16,20 @@ library(lubridate)
...
 
 # ---- Derivations ----
+
+# Derive flags
 adcm <- cm %>%
 
   # Join supplementary qualifier variables
   # derive_vars_suppqual(suppcm)
```
thomas-neitmann, line 26: Please add a trailing `%>%` such that when this is uncommented it actually works.

inst/templates/ad_adcm.R
```diff
@@ -16,16 +16,20 @@ library(lubridate)
...
 
-  # Join adsl to cm
-  left_join(adsl, by = c("STUDYID", "USUBJID")) %>%
+  # Join ADSL with CM (only ADSL vars required for derivations)
+  left_join(
+    adsl %>% select(STUDYID, USUBJID, TRTSDT, TRTEDT, DTHDT, EOSDT),
+    by = c("STUDYID", "USUBJID")
```
thomas-neitmann, line 31: Please don't use nested `%>%` but rather `select(adsl, STUDYID, ...)`.

inst/templates/ad_adcm.R
```diff
@@ -34,7 +38,7 @@ adcm <- cm %>%
     date_imputation = "first",
     time_imputation = "first",
     min_dates = list(TRTSDT)
-  ) %>%
+    ) %>%
```
thomas-neitmann, line 41: The trailing parenthesis should always be aligned with the function name, so

```
derive_vars_dtm(
  date_imputation = "first",
  time_imputation = "first",
  min_dates = list(TRTSDT)
) %>%
```

rather than

```
derive_vars_dtm(
  date_imputation = "first",
  time_imputation = "first",
  min_dates = list(TRTSDT)
  ) %>%
```

thomas-neitmann (changes requested):
Please set `compress = "bzip2"` for all datasets when you use `save()`. Also, we should use lowe/uppercase consistenly for the saved dataset file names. Since we use lowercase for the SDTM datasets I'd stick to that.

Ragarding `AVALU` I would remove that indeed.

There are some lintr comments like this one. You can see them all [here](https://github.com/Roche-GSK/admiral/pull/584/files).
image

bundfussr:
I would agree to removing `AVALU`. It is Roche specific and the ADaMIG states: 
>"Note that there is no separate units variable for BDS variables PARAM or AVAL, since the units of AVAL will be included in the value of PARAM."

I would populate `ADURU` even if `ADURN` is missing. CDISC does not require to set `ADURU` to missing if `ADURN` is missing. I think this makes only sense if the unit depends on the input data and differs across the observations. But this is very unlikely in OCCDS.

inst/templates/ad_adae.R
```diff
@@ -110,6 +116,14 @@ adae <- ae %>%
     mode = "last"
   )
 
+# Join all ADSL with AE
+adae <- adae %>%
+
+  select(-TRTSDT, -TRTEDT, -DTHDT, -EOSDT) %>%
```
kabis-ops, line 122: @thomas-neitmann, @aehmann-gsk  : Should we implement something to avoid this? These adsl variables are added at the start of the pgm to help derivation of other variables and removed at the end before merging again with the full adsl. 
Ideally, we would define only once the variables to add/remove.
We have discussed this already, but not sure this is on the plan yet?

kabis-ops (commented):
Hi @millerg23 
Thanks for the update: here are my comments:
- DTs can now be derived using the derive_vars_dtm_to dt( ) function from the DTM variables. This probably needs to be updated
- the convert_to_na() function is missing from adae/adcm/adex/adsl
- adae/adcm: 
 + derive_vars_duration() and derive_last_dose have a lot of parameters: should the param with a default value be removed from the template?
 + the derive_vars_query()( function is now available), do we want to add an example in the template?

inst/templates/ad_adcm.R
millerg23: Done throughout the scripts.

inst/templates/ad_adcm.R
millerg23: Done.

inst/templates/ad_adae.R
millerg23: For the moment, I have tried to drop the ADSL vars once they have been used. Then add all ADSL vars at the end, this avoids carrying through lots of variables, and having to update them for new records created (long byvar statements).

thomas-neitmann: How about something like this at the start:

```r
adsl_vars <- vars(TRTSDT, TRTEDT, DTHDT, EOSDT)
adae <- ae %>%
  left_join(select(adsl, !!!adsl_vars))
```

and at the end

```r
adae <- adae %>%
  left_join(select(adsl, -!!!adsl_vars))
```

bundfussr: That's the same I had in mind.

For datasets with derived parameter we need to add `adsl_vars` to the `by_vars` parameter.

millerg23:
@kabis-ops I updated adsl/adex/adcm/adae to use derive_vars_dtm_to dt
@thomas-neitmann I update data.R file (thanks!).

inst/templates/ad_adae.R
bundfussr: I have implemented the function `negate_vars()` for it in the `546_derive_param_tte` branch. See https://github.com/Roche-GSK/admiral/blob/546_derive_param_tte/R/utils.R .

millerg23: Have updated all scripts creating VADs with regard to adding ADSL vars, as we discussed.

PR #610, opened 2021-10-19
This is the first draft version of the code.

pk505289:
As per the request i have updated the code. This is the first draft version.


R/derive_summary_records.R
```diff
@@ -196,12 +196,39 @@ derive_summary_records <- function(dataset,
...
-      mutate(!!!set_values_to)
+    get_summary_records(
+      dataset,
+      by_vars = by_vars,
+      filter = !!filter,
+      analysis_var = !!analysis_var,
+      summary_fun = summary_fun,
+      set_values_to = set_values_to)
   )
+  get_summary_records <- function(dataset,
```
bundfussr, line 211: The `get_summary_records()` function should not be defined *within* the `derive_summary_records()` function but as a separate function. Otherwise, you can not call it in `derive_params_exposure()`.

Please add a header to document the `get_summary_records()` function. It could be useful for users. Thus I would export it. For the `@keywords` tag I would use `adam user_utility`.

R/derive_summary_records.R
```diff
@@ -196,12 +196,39 @@ derive_summary_records <- function(dataset,
     subset_ds <- dataset
   }
```
bundfussr, line 197: This part (lines 189-197) can be removed as the filtering is done in `get_summary_records()`.

R/derive_summary_records.R
```diff
@@ -196,12 +196,39 @@ derive_summary_records <- function(dataset,
...
+  } else {
+    subset_ds <- dataset
+  }
+
+  # Summarise the analysis value
+  subset_ds %>%
+    group_by(!!!by_vars) %>%
+    summarise(!!analysis_var := summary_fun(!!analysis_var)) %>%
+    mutate(!!!set_values_to)
+
```
bundfussr, line 232: Please use the simplified version I proposed in my last change request.

R/derive_params_exposure.R
```diff
@@ -166,6 +166,7 @@ derive_params_exposure <- function(dataset,
     ) %>%
     filter(PARAMCD == quo_get_expr(set_values_to$PARAMCD))
```
bundfussr, line 167: `get_summary_records()` should be called here (lines 161-167).

bundfussr (changes requested):

R/derive_summary_records.R
pk505289: Updated the code please check.

R/derive_summary_records.R
pk505289: updated the code. Please check.

R/derive_summary_records.R
pk505289: updated the code. Please check.

bundfussr: The `get_summary_records()` function is still defined within the `derive_summary_records()` function.

R/derive_summary_records.R
bundfussr: Please use
```
  # Summarise the analysis value
  dataset %>%
    group_by(!!!by_vars) %>%
    filter_if(!!filter) %>%
    summarise(!!analysis_var := summary_fun(!!analysis_var)) %>%
    mutate(!!!set_values_to) %>%
    ungroup()
```

R/derive_summary_records.R
bundfussr: The filtering code is still there.

R/derive_summary_records.R
pk505289: separated the code from the main code.

bundfussr:
I just noticed that `filter_if()` does not accept grouped data frames. I would propose to add the parameter `accept_grouped` to `assert_data_frame()`. The default value should be `FALSE`. If it is set to `TRUE`, the check on grouped data frame should be skipped. Then you can set `accept_grouped = TRUE` in the `assert_data_frame()` call in `filter_if()`.

@thomas-neitmann , do you agree?

thomas-neitmann:
Fine with me 👍

R/derive_params_exposure.R
alanaharris22: Thanks @bundfussr, I tried adding that, however when testing I now get an error `Error in is_quosure(arg) : object 'PARAMCD' not found`.  I will continue to look into it, in the meantime I have created a draft PR here - https://github.com/pharmaverse/admiral/pull/1118
Thanks for your help, 
Alana

bundfussr: Please try adding
```
filter <- assert_filter_cond(enquo(filter), optional = TRUE)
```
in `get_summary_records()`. This quotes the `filter` argument and ensures that it is a condition.

alanaharris22: Thanks @bundfussr, I did try adding that, as well as wrapping the filter condition in `rlang::quo()` in `derive_param_exposure` - however I'm now getting a different error when the filtering is applied... I am probably missing something somewhere, will keep looking into it.

bundfussr: In the call of `get_summary_records()` you should not use `quo()`, i.e., it should be `filter = PARAMCD == input_code`.

bundfussr: We plan to add more guidance regarding quosures in the programming strategy.

PR #611, opened 2021-10-19
Fix typo in unit tests

epijim:
See commit message - this appears to the root cause of the validation tables looking bad: https://github.com/insightsengineering/thevalidatoR/issues/32


thomas-neitmann:
Could you merge `devel` into this branch and elete the existing workflow file for the validation report? That has another name so it is not overwritten by your file.

PR #630, opened 2021-10-29
437 update derive var trtsdtm bug

PoojaKumari05:
Hi @teckla-gsk, as discussed I couldn't proceed much but I tried to implement some of the updates required that we discussed yesterday. You can take from here. Thanks for your help.


teckla-gsk:
@bundfussr @kabis-ops I am having a few technical issues with admiral so I have not created the test files but as I explore those technical issues, I would appreciate some feedback on the two files derive_vars_trt_end and derive_vars_trt_start.
Thanks

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+#' @param ord_filter
+#'  The mode by which the filter_extreme is chosen.
+#'  If "first", the first observation of each by group is included in the output dataset.
+#'  If "last", the last observation of each by group is included in the output dataset.
+#'
+#'  Permitted Values: "first", "last"
+#'
+#' @details For each group (with respect to the variables specified for the
+#'   `by_vars` parameter) the first observation (with respect to the order
+#'   specified for the `order` parameter) is included in the output dataset.
```
kabis-ops, line 85: I would suggest something like : For each group (with respect to the variables specified for the `by_vars` parameter), the first or last (with respect to the order specified for the `order` parameter)  `dtc`  is selected and used as input for 'new_var'. We should also mention about the imputation variables.

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+                                 time_imputation,
+                                 min_dates = NULL,
+                                 max_dates = NULL,
+                                 ord_vars,
+                                 ord_filter) {
+    new_var <- assert_symbol(enquo(new_var))
+    dtc <- assert_symbol(enquo(dtc))
+    ord <- assert_symbol(enquo(ord))
+    assert_data_frame(dataset,subject_keys)
+    assert_data_frame(dataset_ex, required_vars = quo_c(subject_keys, ord_vars))
```
kabis-ops, line 129: `subject_keys` is already checked before. The `dtc`  should also be present.

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+    filter_ex <- assert_filter_cond(enquo(filter_ex), optional = TRUE)
+
+    add <- dataset_ex %>%
+      mutate(imputed_dtc = convert_dtc_to_dtm(!!dtc,
+                                              date_imputation = date_imputation,
+                                              time_imputation = time_imputation,
+                                              min_dates = min_dates,
+                                              max_dates = max_dates)
+      )
+    if (!quo_is_null(filter_pre_timepoint)){
```
kabis-ops, line 141: `filter_pre_timepoint` is not a parameter of the function

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+      add %>% filter_if(filter_ex) %>%
+        filter_extreme(
+          order = vars(!!dtc, !!ord_vars),
+          by_vars = !!!subject_keys,
+          mode = ord_filter) %>%
+        mutate(!!new_var := imputed_dtc)
+    } else {
```
kabis-ops, line 148: I dont think we need this else statement: the first block before should be enough?

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+      filter_extreme(
+        order = ord,
+        by_vars = !!!subject_keys,
+        mode = ord_filter) %>%
+        mutate(!!new_var := imputed_dtc)
+    }
+
```
kabis-ops, line 155: We should also add the imputation variable(s) if specified.

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+    dtc <- assert_symbol(enquo(dtc))
+    ord <- assert_symbol(enquo(ord))
+    assert_data_frame(dataset,subject_keys)
+    assert_data_frame(dataset_ex, required_vars = quo_c(subject_keys, ord_vars))
+    assert_character_scalar(ord_filter, values = c("first","last"))
+
+    filter_ex <- assert_filter_cond(enquo(filter_ex), optional = TRUE)
+
+    add <- dataset_ex %>%
+      mutate(imputed_dtc = convert_dtc_to_dtm(!!dtc,
```
kabis-ops, line 135: Do we also want the option to derive a `--SDT`? If not, should we rename the function so that it is clear that it returns the `--SDTM`? Or make it very clear in the documentation?

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+        mutate(!!new_var := imputed_dtc)
+    } else {
+      filter_extreme(
+        order = ord,
+        by_vars = !!!subject_keys,
+        mode = ord_filter) %>%
+        mutate(!!new_var := imputed_dtc)
+    }
+
+    left_join(dataset, add, by = vars2chr(subject_keys)) %>%
```
kabis-ops, line 156: At this stage the `add` datasets will contain all variables from `dataset_ex` and they will be added to the output dataset. Need to select only the relevant variables.

kabis-ops (commented):
Hi @teckla-gsk , Thank you for theupdate :)!
I have reviewed the `derive_vars_trt_start()` function only, but I guess most of the comments apply to both functions. Let me know if anything is not clear or required discussion. Thanks

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+#' @param subject_keys Variables to uniquely identify a subject
+#'
+#' A list of quosures where the expressions are symbols as returned by
+#' `vars()` is expected.
+#'
+#' @param new_var name of treatment start date e.g. `TRTSDTM`
+#'
+#' @param dtc name of variable from which start date is derived e.g. `EXSTDTC`
+#'
+#' @param date_imputation The value to impute the day/month when a datepart is
```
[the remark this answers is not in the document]
teckla-gsk: @kabis-ops  I am not sure what you mean by this

R/derive_vars_trt_start.R
teckla-gsk: @kabis-ops add the imputed date as a separate variable or replace the partial date? Also how do add the flag, as a separate variable?

R/derive_vars_trt_start.R
teckla-gsk: I believe the original was --DTM?

R/derive_vars_trt_start.R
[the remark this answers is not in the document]
kabis-ops: Hi, if you use the `@inheritParams impute_dtc` then you should automatically get the same documentation as in the `impute_dtc()` function without repeating them (and hen we make sure they are consistent). There is an example in the convert_dtc_to_dt() header.

R/derive_vars_trt_start.R
kabis-ops: Sorry, i meant the imputation flags.

R/derive_vars_trt_start.R
kabis-ops: Yes, I agree. So we should make it clear in the documentation because it is not obvious from the function name (i think ;0). 
Also, we should probably add a warning if the user enter a `new_var` like `TRTSDT` rather than `TRTSDTM`?

teckla-gsk:
@kabis-ops @bundfussr please review the scripts and the test files for derive_vars_trt_start/end I will focus on this tomorrow morning so we can get it done :) Thanks

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,152 @@
...
+    msg <- sprintf(
+      "Input EX dataset has not been filtered. Check study requirement if ex needs to be filtered.")  # nolint
+    warn(msg)
+  } else {
+    add <- add %>% filter(!!filter_ex)
+  }
+
+  #pick last or first date in sequence
+  add <- add %>%
+    filter_extreme(order = order,
```
kabis-ops, line 117: the imputed date should be used in the order.

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,152 @@
...
+                   by_vars = subject_keys,
+                   mode = ord_filter) %>%
+    mutate(!!new_var := imputed_dtc) %>%
```
kabis-ops, line 120: all the code below (up to line 149) should not be needed if `derive_vars_dtm()` is used.

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,152 @@
...
+                                  dataset_ex,
+                                  filter_ex = NULL,
+                                  subject_keys,
+                                  new_var,
+                                  dtc,
+                                  date_imputation,
+                                  time_imputation,
+                                  min_dates = NULL,
+                                  max_dates = NULL,
+                                  flag_imputation = "AUTO",
```
kabis-ops, line 75: Could you please move `flag_imputation` just after `time_imputation` for consistency with other functions?

R/derive_vars_trt_end.R
```diff
@@ -0,0 +1,152 @@
...
+
+  #filter EX data based on study requirements
+  if (quo_is_null(filter_ex)) {
+    add
+    msg <- sprintf(
+      "Input EX dataset has no been filtered. Check study requirement if ex needs to be filtered.")  # nolint
+    warn(msg)
+  } else {
+    add <- add %>% filter(!!filter_ex)
+  }
```
bundfussr, line 114: I would remove the warning because if `filter_ex` is null, it was specified by the user. See my comment above regarding the default value.

I would apply the filter before the imputation by calling `filter_if()`. This is more efficient.

R/derive_vars_trt_end.R
```diff
@@ -0,0 +1,152 @@
+#' Derive Datetime of Any Treatment End Date
+#'
+#' Derives datetime of any last exposure to treatment e.g. (`TRTEDTM`)
+#'
+#' @param dataset Input dataset
+#'
+#'   The variables specified by the `by_vars` parameter are expected.
```
bundfussr, line 7: "by_vars" should be replaced with "subject_keys"

bundfussr (changes requested):

R/derive_vars_trt_start.R
```diff
@@ -0,0 +1,158 @@
...
+#' any of the specified dates, e.g., that the imputed date is not after the data
+#' cut off date. Only dates which are in the range of possible dates are
+#' considered.
+#'
+#' @param ord_vars
+#'  Order of variables for which either the last or first observation
+#'  is picked as treatment start date
+#'  e.g. vars(EXSTDTC,EXSEQ)
+#'
+#' @param ord_filter
```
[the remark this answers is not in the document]
bundfussr: I agree that for treatment start date we need `mode = "first"` only.

However, at the moment the code of both functions is the same. Thus I wonder if we need _two_ functions. We could keep the `mode` parameter and just implement one function.

What do you think?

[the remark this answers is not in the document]
teckla-gsk: I thought about that when I was writing this, if we go with one function what should be the name? then put in documentation that start/end is controlled by the mode parameter?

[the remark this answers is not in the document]
thomas-neitmann: Maybe `derive_trt_dt()`?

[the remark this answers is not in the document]
kabis-ops: What about :`derive_vars_trt_date()` and 2 wrapper functions :   `derive_vars_trt_start` with mode="first", and `derive_vars_trt_end` would mode="last"?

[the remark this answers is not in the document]
bundfussr: Maybe we should drop the "trt" part because the function is no longer specific for treatment dates. The only reference to treatment is the parameter name `dataset_ex` but the function would work with any dataset.

What about naming the function `derive_vars_extreme_date()` and renaming `dataset_ex` to `dataset_date`?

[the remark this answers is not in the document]
thomas-neitmann: In favor 👍

teckla-gsk:
@thomas-neitmann @kabis-ops @bundfussr 
would you please review the combined function? I believe the agreement is to go with a combined function, also maybe in the review we can consider the implications of how many dependencies admiral functions have

R/derive_vars_extreme_date.R
```diff
@@ -0,0 +1,154 @@
...
+#'                           max_dates = NULL,
+#'                           mode="LAST") %>%
+#'   select(USUBJID, TRTEDTM)
+derive_vars_extreme_date <- function(dataset,
+                                     subject_keys =vars(STUDYID,USUBJID),
+                                     dataset_date,
+                                     filter_dataset_date = NULL,
+                                     dtc,
+                                     order,
+                                     new_var,
```
bundfussr, line 73: Should we replace `new_var` with `new_var_prefix`?

R/derive_vars_extreme_date.R
```diff
@@ -0,0 +1,154 @@
...
+                   by_vars = subject_keys,
+                   mode = mode) %>%
+    mutate(!!new_var := imputed_dtc) %>%
+    select(!!!subject_keys, !!new_var, !!dtc, imputed_dtc)
+
+  #add imputation flags
+  if (flag_imputation %in% c("both", "date") ||
+      flag_imputation == "auto" && !is.null(date_imputation)) {
+    # add --DTF if not there already
+    dtf <- paste0("TRT", "DTF")
```
bundfussr, line 129: The name of the date imputation flag should not be hard-coded.

R/derive_vars_extreme_date.R
```diff
@@ -0,0 +1,154 @@
...
+  new_var <- assert_symbol(enquo(new_var))
+  dtc <- assert_symbol(enquo(dtc))
+  assert_data_frame(dataset, subject_keys)
+  assert_data_frame(dataset_date, required_vars = quo_c(subject_keys, order))
+  assert_character_scalar(mode, values = c("first", "last"), case_sensitive = F)
+
+  filter_dataset_date <- assert_filter_cond(enquo(filter_dataset_date), optional = TRUE)
+  filter_dataset_date <- enquo(filter_dataset_date)
+
+  assert_character_scalar(flag_imputation, values = c("auto", "both", "date", "time", "none"),
```
bundfussr, line 90: To make `flag_imputation` case insensitive `flag_imputation <- assert_character_scalar(...` should be used. (`assert_character_scalar()` returns the lower case value if `case_sensitive = FALSE` is specified.)

R/derive_vars_extreme_date.R
```diff
@@ -0,0 +1,154 @@
...
+                                     mode) {
+
+  new_var <- assert_symbol(enquo(new_var))
+  dtc <- assert_symbol(enquo(dtc))
+  assert_data_frame(dataset, subject_keys)
+  assert_data_frame(dataset_date, required_vars = quo_c(subject_keys, order))
+  assert_character_scalar(mode, values = c("first", "last"), case_sensitive = F)
+
+  filter_dataset_date <- assert_filter_cond(enquo(filter_dataset_date), optional = TRUE)
+  filter_dataset_date <- enquo(filter_dataset_date)
```
bundfussr, line 88: Should be removed as the value is already quoted in the previous line.

R/derive_vars_extreme_date.R
```diff
@@ -0,0 +1,154 @@
...
+
+  #Issue warning that new_var should be DTM
+  if (str_detect(rlang::quo_name(new_var), "DTM", negate = T)) {
+    msg <- sprintf(
+      "This function derives the treatment datetime, new_var must end in DTM.")
+    warn(msg)
+  }
+
+  add <- dataset_date %>%
+    mutate(imputed_dtc = convert_dtc_to_dtm(!!dtc,
```
bundfussr, line 101: Couldn't we use `derive_vars_dtm()` instead of `mutate()` and `convert_dtc_to_dtm()`?
Then we could also remove the code for deriving the imputation flags below.

bundfussr (changes requested):

R/derive_vars_extreme_date.R
teckla-gsk: Done

R/derive_vars_extreme_date.R
teckla-gsk: good catch

R/derive_vars_extreme_date.R
teckla-gsk: ok

R/derive_vars_extreme_date.R
teckla-gsk: I would rather not since derive_vars_dtm uses the same mutate and convert_dtc_to_dtm meaning we would be nesting dependencies.

teckla-gsk:
@thomas-neitmann @bundfussr @kabis-ops updated based on comments above. I thought to separate to extreme_dtm and extreme_dt since this looks like the strategy for a lot of the other date functions, however if one runs any _dtm function they could probably just run mutate and as.date on the created -dtm variable, but I think the point here is to provide a full suite? Anyway, looking forward to review comments

teckla-gsk:
@thomas-neitmann @bundfussr @kabis-ops @aehmann-gsk  there is a piece of user feedback that suggest this functions (extracting trtsdt) should allow for those studies that check multiple dates e.g. use earliest date among exendt, dthdt and trial enddate. I don't think we are implementing this now but a valid point on extreme dates

R/derive_vars_extreme_date.R
bundfussr: I think if the code should do the same as the code of `derive_vars_dtm()`, we should not copy the code but call `derive_vars_dtm()`. Then
- it is clear from the code that the standard derivation of `derive_vars_dtm()` is used here,
- it ensures a consistent behaviour of the functions, and
- it is easier to maintain.

bundfussr:
I am not sure what you mean by "provide a full suite". I think providing `derive_vars_extreme_dtm()` and `derive_vars_extreme_dt()` is in line with our current strategy. As you already mentioned `derive_vars_extreme_dt()` could be implemented by calling `derive_vars_extreme_dtm()` and `mutate()`. I would prefer this as it avoids duplication of code.

The user feedback looks more like a request to extend `derive_var_lstalvdt()`.

aehmann-gsk:
I think the user is probably needing something similar to below? My team plans to generalize and put into our TA package as we are frequently trying to find the maximum date not only based on a series of records but also based on a series of dates within a record.

```
  temp_adsl <- select(adsl, STUDYID, USUBJID, DOSEDT1, DOSEDT2, DOSEDT3, DOSEDT4) %>% 
    rename_with(~ gsub("DT", "DT_TEMP", .x, fixed = TRUE)) %>% 
    rowwise() %>%
    mutate(FU_TEMP1 = ifelse(!is.na(DOSEDT2_TEMP) | !is.na(DOSEDT3_TEMP) | !is.na(DOSEDT4_TEMP),
                             max(DOSEDT2_TEMP + 67, DOSEDT3_TEMP + 1, DOSEDT4_TEMP + 1, na.rm=TRUE),
                             NA_Date_),
           FU_TEMP = ifelse(!is.na(FU_TEMP1) | !is.na(DOSEDT1_TEMP),
                            min(DOSEDT1_TEMP, FU_TEMP1, na.rm = TRUE),
                            NA_Date_),
           LAST_TEMP = ifelse(!is.na(DOSEDT2_TEMP) | !is.na(DOSEDT3_TEMP),
                              max(DOSEDT2_TEMP, DOSEDT3_TEMP, na.rm=TRUE),
                              NA_Date_)) %>% 
    ungroup()
```
(Just a note - I pulled this snippet of code from a study function and tried to dummy it quickly from a function we have but it gives the idea.)

Once thing about the `rowwise()` computations, you have to be extremely careful with not creating `-Inf` as a result as it appears to the user as an `NA` in their output but is not `NA` and can make subsequent code not work appropriately if the user is testing based on `NA`.  That is one reason it would be worthwhile to have a function in admiral as to handle the `-Inf` on rowwise() computations.  At the very least, it may be useful to include such an example in the `ADSL` of `ADEX` vignettes.

bundfussr:
@aehmann-gsk , I think most of the functionality is already provided by `derive_var_lstalvdt()`. (You can use the same dataset for different `lstalvdt_source()` objects.) Missing is the `new_var` parameter and the option to add a certain number of days to a date like in your example.

teckla-gsk:
@thomas-neitmann @bundfussr I suppose then it comes down to; do we stick with the original strategy of variable specific functions and revert to the original derive_var_trtst(ed) and update the lstalvdt function to be the extreme date or do we move toward a "use-case" strategy for functions to which I agree with Stefan that all these could be handled in the generalization of lstalvdt ? Then the wider question on what strategy to use moving forward especially in responding to feedback?

aehmann-gsk:
I think a discussion is warranted on this because this sort of logic is used extensively in APHASE, AVISIT, etc. logic.  Doing the `lstalvdt` with different limits based on visits, phases, etc.. could introduce a lot of overhead that is unwarranted in this sort of processing.

thomas-neitmann:
Closing this as it's been stale for months.

PR #632, opened 2021-11-01
Make unit check case insensitive

### LINKED ISSUE #631: Unit checking should be case INsensitive
Currently checking for the proper units in `derive_param_*()` functions is case sensitive which leads to annoying errors such as this one.

```
Error: It is expected that HR is measured in beats/min.
In the input dataset it is measured in `BEATS/MIN`
```

Let's make the check case insensitive and ensure that these kind of error do *not* appear.

thomas-neitmann:
Close #631 

@Roche-GSK/admiral-devs Would be great if someone could review this today as I'd like to have this functionality included for the R in Pharma demo tomorrow.

### DIFF
--- a/R/assertions.R
+++ b/R/assertions.R
@@ -860,16 +860,16 @@ assert_unit <- function(dataset, param, required_unit, get_unit_expr) {
     pull(`_unit`) %>%
     unique()
 
-  if (length(units) != 1L || units != required_unit) {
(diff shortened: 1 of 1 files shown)


malexthorpe (approved):
All looks good.

thomas-neitmann:
Thanks @malexthorpe!

PR #640, opened 2021-11-05
Move SDTM data to cdiscpilot pkg

### LINKED ISSUE #639: Move SDTM datasets into their own package
Having the SDTM datasets as part of admiral causes some trouble. For one we get a NOTE when running `R CMD check` saying the package size is too large and also running both `devtools::document()` and `devtools::check()` seems to be extremely slowed down by the presence of all the datasets.

thomas-neitmann:
Close #639


thomas-neitmann:
@bundfussr @aehmann-gsk Woud be great if you could approve this PR.

millerg23 (approved):
All looks good, no comments.

PR #675, opened 2021-11-18
572 imputation: update imputation vignette

bundfussr:


malexthorpe (approved):
All looks good and conforms to the writing vignettes guidance.

PR #748, opened 2021-12-09
Test that template scripts run w/o errors

### LINKED ISSUE #730: VAD script execution should be part of CI/CD
Scripts that create VADs should be run automatically, if any functions used in script have been updated. As well as checking for ERRORs, should we check for differences between versions of VAD?
(assigned to Thomas initially to get prioritised in Sprints!)

thomas-neitmann:
Close #730


thomas-neitmann:
I initially created a unit test file for running all templates and checking whether any error occur. However, given the overhead of running these template scripts I do not want to run them on every single commit to a PR. Thus, instead I've created a new GitHub actions workflow that run whenever something is ushed to `deve` or `master`, i.e. when a PR is merged, or alternatively whenever it is manually triggered.

bundfussr:
Maybe we could add it to the review process. I.e., before the reviewer is merging the PR she/he should run the checks manually.

bms63 (commented):
So this identifies the template that is the culprit in the github action?  Can we test it out with a dummy branch?  These Github Actions are amazing

thomas-neitmann:
> Maybe we could add it to the review process. I.e., before the reviewer is merging the PR she/he should run the checks manually.

@bundfussr I really want to add another manual review step. We already have a lot of these.

I just found out that you can trigger a workflow on PR approval [here](https://github.community/t/feature-request-trigger-action-on-pull-request-approved/18413/2). That looks promosing, don't you think?

> So this identifies the template that is the culprit in the github action? Can we test it out with a dummy branch? These Github Actions are amazing

Yes, exactly. It runs all templates, looks which ones errored and then prints a list of those. I added manual workflow triggering as an option but this seems to only be possible *after* this branch has been merged.

bundfussr:
Yes, triggering the workflow on PR approval is a much better solution.

bms63 (approved):
That .yml file is pretty amazing

PR #750, opened 2021-12-10
Deprecate `derive_baseline()` and `derive_var_basec()`

### LINKED ISSUE #695: Rethink interface of baseline related derivations
Several tester indicated that they found the naming of `derive_baseline()` to be confusing as they expected it to derive the baseline flag. In addition, questions where raised whether there's a need to have `derive_var_base()` and `derive_var_basec()` if you can use the underlying function to create those witht the added benefit of more explicit code as you would have to specify the `source_var` and `new_var` when using `derive_baseline()`.

My suggestion would be to rename `derive_baseline()` to `derive_var_base()` and remove `derive_var_basec()`.

thomas-neitmann:
Close #695


R/derive_var_base.R
```diff
@@ -0,0 +1,130 @@
...
+  source_var <- assert_symbol(enquo(source_var))
+  new_var <- assert_symbol(enquo(new_var))
+  assert_data_frame(
+    dataset,
+    required_vars = quo_c(by_vars, source_var, quo(ABLFL))
+  )
+  warn_if_vars_exist(dataset, quo_text(new_var))
+
+  base <- dataset %>%
+    filter(ABLFL == "Y") %>%
```
bundfussr, line 83: The condition should not be hard-coded.

bundfussr (changes requested):
Please update
- the BDS vignette,
- the ADEG and ADVS script, and
- the changelog

NEWS.md
```diff
@@ -1,3 +1,7 @@
+# admiral development version
```
bundfussr, line 1: Why did you use "development version" instead of "0.6.0"?

It seems that it prevents the section to be displayed. Is this intended?

thomas-neitmann: I don't think it's set in stone that the next version will be 0.6.0 so I used "development version" instead. That shouldn't affect the rendering, though. But even if it does, this will obviously be replaced by an actual version number in the next release (and only then will it be displayed on our website).

bundfussr (approved):

Issue #722, opened 2021-12-01
Correct example for `derive_params_exposure()`


bundfussr:
There is a typo in the `filter` parameter of the second example ("average dose in w2-24"): "WEEK2" instead of "WEEK 2"

The values of `USUBJID` in the examples should be shortened such that the subjects are distinguishable, e.g., by removing the "01-701-" part.

thomas-neitmann:
Btw, why is the function call derive_param**s**_exposure() and not derive_param_exposure? If I correctly remember our discussion it only adds a single parameter per call. If so, then we should rename it accordingly.

bundfussr:
I agree that we should remove the "s". I think the only function which could add more than one parameter is `derive_param_tte()` and this is without "s".

rossfarrugia:
@thomas-neitmann @bundfussr i'd like to test out our contribution model putting myself in the shoes of an external contributor (well technically i am having contributed zero admiral code thus far!), so is it OK to assign this issue to me? no problem if someone else is already picking it up

bundfussr:
@rossfarrugia , I do not think that anyone is working on this already. So feel free to assign it to you.

rossfarrugia:
thanks @bundfussr i will pick it up. @thomas-neitmann i don't have permission to assign people to admiral issues so please assign me

thomas-neitmann:
@rossfarrugia Great initiative 👏 

I just gave to "write" access to the repository. You should now be able to assign yourself (or anyone else for that matter) and push new branches to the repo. Good luck 😉

rossfarrugia:
@bundfussr i'm shamed by how long that took me this morning, but good learning nonetheless! i updated the derive_params_exposure R program to rename to param instead of params and updated the examples, and i updated the corresponding unit test program with params->param. see the pull request linked above. shout if i missed anything!

@thomas-neitmann this [page](https://roche-gsk.github.io/admiral/articles/git_usage.html) is excellent!

PR #760, opened 2021-12-16
722 derive param exposure

rossfarrugia:
#722 i updated the derive_params_exposure R program to rename to param instead of params and updated the examples, and i updated the corresponding unit test program with params->param. hope i didn't miss anything! at first i did this via RStudio and created new versions of the programs with the new name and pushed them here, but then realised i could do such minor updates way easier in GitHub directly so i deleted those new file commits and just updated and renamed the 2 "params" files directly.


bundfussr (changes requested):
Thanks for your contribution.

Could you please
- link the issue to the pull request,
- deprecate the old function, i.e., `derive_params_exposure()` should still work but issue a warning, see the "Deprecation" section in the [Programming Strategy](https://roche-gsk.github.io/admiral/articles/programming_strategy.html#deprecation-1) for details and derive_var_base.R in the devel branch for an example, and 
- update the Changelog (NEWS.md)?

rossfarrugia:
@bundfussr i think i've done all those pieces now - can you take another look please? thanks for all the tips!

bundfussr (changes requested):
Please add
```
deprecate_warn("0.6.0", "derive_params_exposure()", "derive_param_exposure()")
```
as first line in `derive_params_exposure()`. This issues a warning if the function is called.

Could you also move line 54/55
```
#'   *Permitted Values:* List of variable-value pairs` to line
#'
```
to line 50? It is not related to the issue. Just a typo we did not notice before.

rossfarrugia:
got it @bundfussr - now updated

bundfussr (changes requested):
Thanks, could you rerun `devtools::document()` and add the updated files in man/?

rossfarrugia:
@bundfussr yes sorry missed that, now done.

bundfussr (approved):

thomas-neitmann:
Thanks a lot @rossfarrugia! I spotted two issues (unfortunately only after this PR has been merged):

1. The code of the deprecated `derive_params_exposure()` is still present in the function body. After calling `deprecate_warn()` it should simply pass its input arguments to the "new" `derive_param_exposure()`. Here's an example: https://github.com/Roche-GSK/admiral/blob/32c8b1704b3dc16e66a0ebaaab1c8880e6330667/R/derive_var_base.R#L132-L140
2. There should be a test case in `tests/testthat/deprecation.R` that checks whether a warning is issued when using the old `derive_params_exposure()`. Something like this: https://github.com/Roche-GSK/admiral/blob/32c8b1704b3dc16e66a0ebaaab1c8880e6330667/tests/testthat/test-deprecation.R#L216-L237

Is this something you could add and open a new PR? If not, I'm happy to take care of it.

rossfarrugia:
i see @thomas-neitmann - i can update this no problem

bundfussr:
Sorry, I missed those.

Could we add the two points to the "Deprecation" section in the "Programming Strategy"?

thomas-neitmann:
We should, yes. Let's open a new ssue for that.

bundfussr:
@rossfarrugia has added them to https://github.com/Roche-GSK/admiral/issues/765 👍 .

PR #702, opened 2021-11-29
Update derive_vars_dtm to change min_dates with max_dates for lapply …

hamzarahal:
Update derive_vars_dtm to put max_dates instead of min_dates when calling lapply
Update the NEWS to describe the bug
Add unit test to cover the case

PR #762, opened 2021-12-16
698 add reference to list all templates

hamzarahal:


bms63:
Hi @hamzarahal Hamza.  Just seeing where we are at with this PR?

amsmith214 (dismissed):
think i've got this in good shape

amsmith214 (commented):
on further review, need to try and bridge this PR and #702

amsmith214:
kinda stumped as to why this is failing on release

bms63:
> kinda stumped as to why this is failing on release

Stumped on why the test is failing?

amsmith214:
> > kinda stumped as to why this is failing on release
> 
> Stumped on why the test is failing?

yeah. it passes 3.6, but fails on release, seemingly because of some differing column attributes, and i'm not sure how to explicitly control that, or where the difference is being introduced, especially as i'm using essentially the same code that passes everywhere else in the test files

amsmith214:
thanks @bms63 !!!!

wqrena:
@amsmith214 the unit test for max_dates looks good to me.

thomas-neitmann:
This contains unrelated changes for `derive_vars_dtm()`. I would suggest to create a fresh PR that only include the updates to the documentation of the template functions.

thomas-neitmann:
Please ignore my previous comment.

R/user_helpers.R
```diff
@@ -1,5 +1,7 @@
 #' Open an ADaM Template Script
 #'
+#' Uses list_all_templates() to discover which templates are available
```
thomas-neitmann, line 3: Please move this to the `@details` section. Also please put the function name into backticks. Plus it should not be Use~s~ but just Use.

thomas-neitmann (changes requested):

R/user_helpers.R
amsmith214: done

thomas-neitmann (approved):

PR #766, opened 2021-12-17
722 derive params exposure deprecation fix

rossfarrugia:
Attempted updates for @thomas-neitmann's spots at #760


thomas-neitmann (approved):
Thank you!

thomas-neitmann:
So I finally managed to make the "Check Templates" workflow working and sure enough it errored because some updates we've made where not incorporated into the template scripts. This is not specific to this PR, though. I'll adjust the templates in a separate PR.

PR #778, opened 2021-12-20
Create workflow to update docs on PR approval

### LINKED ISSUE #775: Create workflow to automatically create man files
The workflow should be triggered whenever something is pushed to `devel` or `master`, run `devtools::document()` and commited any updated file in the `man` folder.

thomas-neitmann:
Close #775

### DIFF
--- a/.github/workflows/update_documentation.yml
+++ b/.github/workflows/update_documentation.yml
@@ -45,11 +45,16 @@ jobs:
           renv::restore()
         shell: Rscript {0}
 
+      - name: Extract branch name
+        id: extract_branch
+        run: echo "##[set-output name=branch;]$(echo ${GITHUB_REF#refs/heads/})"
+
       - name: Update documentation
         run: |
           Rscript -e "devtools::document()"
           git config --local user.email "actions@github.com"
           git config --local user.name "GitHub Actions"
+          git checkout ${{ steps.extract_branch.outputs.branch }}
           git add --force NAMESPACE man/*.Rd
           git commit --allow-empty -m "Update documentation"
           git push
--- a/.github/workflows/update_documentation.yml
+++ b/.github/workflows/update_documentation.yml
@@ -1,0 +1,56 @@
+on:
+  workflow_dispatch:
+  pull_request_review:
+    types: [submitted]
+
+name: update_documentation
+
+jobs:
+  update_documentation:
+    if: github.event.review.state == 'approved'
+    runs-on: ubuntu-20.04
+    env:
+      GITHUB_PAT: ${{ secrets.GITHUB_TOKEN }}
+    steps:
+      - uses: actions/cache@v2
+        with:
+          path: ~/.local/share/renv
+          key: ${{ runner.os }}-renv-${{ hashFiles('**/renv.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-renv-
+
+      - uses: actions/checkout@v2
+
+      - uses: r-lib/actions/setup-r@v1
+        with:
+          r-version: 3.6
+
+      - name: Query dependencies
+        run: |
+          install.packages('remotes')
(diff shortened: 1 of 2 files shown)


thomas-neitmann:
I'm not 100% sure this works but as I don't know a way of checking GItHub action locally (do you?) I think we have to merge and see what happens the next time a PR is approved.

.github/workflows/update_documentation.yml
```diff
@@ -0,0 +1,56 @@
...
+          if (!requireNamespace("renv", quietly = TRUE)) install.packages("renv")
+          renv::restore()
+        shell: Rscript {0}
+
+      - name: Update documentation
+        run: |
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          Rscript -e "devtools::document()"
+          git add .
```
bundfussr, line 54: I think this adds only the files in the current directory. `git add -A` should add all files.
However, maybe we should use `git add -f NAMESPACE man/*.Rd` and add `NAMESPACE` and `man` to the .gitgnore file.

bundfussr (changes requested):

.github/workflows/update_documentation.yml
thomas-neitmann: Do you think adding to `.gitignore` is necessary? After all, the reason for adding this workflow is that people forget to run `devtools::document()` and commit the created files. I agree on specifically adding only `NAMESPACE` and `man/*.Rd`, though.

bundfussr: I think adding the files to .gitignore makes the development easier. Otherwise, the files are listed as modified, which may cause confusion.
What is the advantage of not adding them to .gitignore?

thomas-neitmann: Fair enough, I've made the change.

bundfussr: What about `NAMESPACE`? Shouldn't it be added to .gitignore as well?

thomas-neitmann: Indeed,

bundfussr (approved):

.github/workflows/update_documentation.yml
```diff
@@ -0,0 +1,55 @@
...
+          if (!requireNamespace("renv", quietly = TRUE)) install.packages("renv")
+          renv::restore()
+        shell: Rscript {0}
+
+      - name: Update documentation
+        run: |
+          Rscript -e "devtools::document()"
+          git config --local user.email "actions@github.com"
+          git config --local user.name "GitHub Actions"
+          git add --force NAMESPACE man/*.Rd
```
bundfussr, line 53: It seems that we need to checkout the feature branch first.

bundfussr (changes requested):

thomas-neitmann:
@bundfussr The workflow failed after it was successfully triggered by your approval. I've just pushed some changes that hopefully make it work. Please approve again such that we can see whether it actually does.

bundfussr (approved):

thomas-neitmann:
Sorry I only pushed now...

bundfussr (approved):

thomas-neitmann:
That didn't work out either, maybe this time...

bundfussr (approved):

thomas-neitmann:
@bundfussr I experimented some more with this and actually maanged to successfully commit and push changes [here](https://github.com/Roche-GSK/admiral/runs/4772801101?check_suite_focus=true). In theory this should work now so please approve again.

bundfussr (approved):

bundfussr:
@thomas-neitmann , I approved but unfortunately it failed again.

bms63:
Can we close this PR - looks like some sort of test you guys were running?

thomas-neitmann:
Yes, let's close. These are my failed attempts at making this work. I'm counting on @dinakar29 to implement something like this for us.

dinakar29:
Hi all, I'll implement this as a workflow😄

PR #779, opened 2021-12-20
#660 add single dose function

### LINKED ISSUE #660: Create function to expand aggregate dosing in EX to single doese per row dataset
Reviewers: Stefan
We need a function that transforms an EX dataset that has multiple doses per row, e.g. with `EXDOSFRQ == "Q2W"`, into one that has only a single dose per row, i.e. `EXDOSFRQ == "ONCE"`.

Example:

**Source**

USUBJID | ASTDT | AENDT | EXDOSFRQ | EXDOSE
--------- | -------- | ------- | ------------ | --------
P01 | 2021-01-01 | 2021-01-07 | Q2D | 100
P01 | 2021-01-08 | 2021-01-15 | Q2D | 100

**Result**

USUBJID | ASTDT | AENDT | EXDOSFRQ | EXDOSE
--------- | -------- | ------- | ------------ | --------
P01 | 2021-01-01 | 2021-01-01 | ONCE | 100
P01 | 2021-01-03 | 2021-01-03 | ONCE | 100
P01 | 2021-01-05 | 2021-01-05 | ONCE | 100
P01 | 2021-01-07 | 2021-01-07 | ONCE | 100
P01 | 2021-01-08 | 2021-01-08 | ONCE | 100
P01 | 2021-01-10 | 2021-01-10 | ONCE | 100

malexthorpe:
Closes #660


malexthorpe:
Hey @thomas-neitmann and @kabis-ops,
The function currently only works for Days and Weeks in a certain format. I looked into making it more robust but there were lots of possible formats for `EXDOSFRQ` (like "10 DAYS PER MONTH" or "THRICE"), so I thought it best to keep it simpler for now.

thomas-neitmann:
Please review @kabis-ops and @konstantinakoukR.

amsmith214:
How large a lookup table do we want to use for this, and where are lookup tables generally stored in the package architecture? 

The lookup table I could make from the list i found at https://evs.nci.nih.gov/ftp1/CDISC/SDTM/CDASH%20Terminology.pdf is relatively short, such that even the case of the example in the issue thread is not covered.

bundfussr:
@amsmith214 , I think you opened the wrong file. It should be https://evs.nci.nih.gov/ftp1/CDISC/SDTM/SDTM%20Terminology.pdf. If you search for "C71113", you'll find all frequencies.

If you create the look-up table in open code instead of within a function, it is added to the package workspace when loading the package. See `tte_source.R` for and example.

amsmith214:
For frequencies like "6 TIMES EVERY MONTH", shall I just assume for the default lookup that doses occur at the closest approximation to "evenly spaced over the given interval" as I can get? E.g., assuming a month is 30 days, doses would happen every 5 days.

bundfussr:
@amsmith214 , yes, I think that makes sense. I just wonder which length of a month we should use. At Roche a month has 30.4375 days and a year has 365.25 days. What is used at GSK?

amsmith214:
> @amsmith214 , yes, I think that makes sense. I just wonder which length of a month we should use. At Roche a month has 30.4375 days and a year has 365.25 days. What is used at GSK?

I actually don't know. I will code using those numbers currently and try to find out what's used here. In previous CRO work I know we used 365.25 for a year, which would indeed imply months have 30.4375 days for these purposes.

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+
+    # Set up lookup table to be joined to dataset
+
+    lookup <- lookup_table %>%
+      rename(!!dose_freq := !!lookup_column)
+
+    # Use compute_duration to determine the number of completed dose periods
+
+    dataset <- dataset %>%
+      left_join(lookup, by = as_name(dose_freq)) %>%
```
amsmith214, line 244: Think I'm in a bit of a bind unless folks have sharp ideas: `as_name` is not in the `rlang` version for R 3.5 it looks like? But if I change it to `quo_name`, I believe that gets me a warning, and going straight to `as.character()[2]` also makes R mad.

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#'
+#' @examples
+#' data <- tibble::tribble(
+#'   ~USUBJID, ~EXDOSFRQ, ~ASTDT, ~AENDT,
+#'   "P01", "Q2D", lubridate::ymd("2021-01-01"), lubridate::ymd("2021-01-07"),
+#'   "P01", "Q3D", lubridate::ymd("2021-01-08"), lubridate::ymd("2021-01-15"),
+#'   "P01", "EVERY 2 WEEKS", lubridate::ymd("2021-01-15"), lubridate::ymd("2021-01-29")
+#' )
+#'
+#' create_single_dose_dataset(data, by_vars = vars(USUBJID))
```
bundfussr, line 218: An example with custom lookup table would be helpful.

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#' @details Each aggregate dose row is split into multiple rows which each represent
+#' a single dose.The number of completed dose periods between `start_date` and
+#' `end_date` is calculated with `compute_duration` and multiplied by the count
+#' of doses per period. `start_date` and `end_date` are then adjusted by an
+#' appropriate increment using `DayConversionFactor`.
+#'
+#' @author Michael Thorpe, Andrew Smith
+#'
+#' @return The input dataset with a single dose per row.
+#'
```
bundfussr, line 207: The `@keywords` tag is missing. I would suggest `@keywords adae adex user_utility`.

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#' @keywords dose_freq
+#'
+#' @rdname dose_freq_lookup
+#'
+dose_freq_lookup <- tibble::tribble(
+  ~NCICode, ~CDISCSubmissionValue,
+  "C64526", "1 TIME PER WEEK",
+  "C139179", "10 DAYS PER MONTH",
+#  "C176288", "2 TIMES PER CYCLE",
```
bundfussr, line 24: What about the out commented lines? Shouldn't we remove them or define values for them?

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#' appropriate functions from `lubridate`.
+#'
+#' @details
+#' To see the entire table in the console, run `print(dose_freq_lookup)`.
+#'
+#' @seealso [create_single_dose_dataset()]
+#'
+#' @export
+#'
+#' @keywords dose_freq
```
bundfussr, line 16: The `dose_freq` keyword is not used on the reference page, i.e., the documentation can not be reached on the website. I would propose to use `@keywords metadata` and add a corresponding subsection to the "Other Keywords" section of the reference page.

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#'
+#' @rdname dose_freq_lookup
+#'
+dose_freq_lookup <- tibble::tribble(
+  ~NCICode, ~CDISCSubmissionValue,
```
bundfussr, line 21: @thomas-neitmann , @bms63 , what is our convention for dataset variable names? Should they be uppercase or does this apply to ADaM datasets only?

bundfussr (changes requested):
Could you add references to the `create_single_dose_dataset()` function?

- In the ADEX vignette
- In the see also section of the last dose functions
- In the description of the `single_dose_condition` parameter of the last dose functions

R/create_single_dose_dataset.R
amsmith214: i was aiming for a lean working version of this that covers most cases. cycles can be of variable length, so i commented them out and did not define values for them. i think i would vote to just remove them if we don't like having commented lines

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
+#' Pre-Defined Dose Frequencies
+#'
+#' These pre-defined dose frequencies are sourced from
+#' [CDISC](https://evs.nci.nih.gov/ftp1/CDISC/SDTM/SDTM%20Terminology.pdf). The
+#' number of rows to generate using `create_single_dose_dataset()` arguments
+#' `start_date` and `end_date` is derived from `DFFactor` and `DFUnit` with
```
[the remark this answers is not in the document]
amsmith214: forgot to change this. will update

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#' data <- tibble::tribble(
+#'   ~USUBJID, ~EXDOSFRQ, ~ASTDT, ~AENDT,
+#'   "P01", "Q2D", lubridate::ymd("2021-01-01"), lubridate::ymd("2021-01-07"),
+#'   "P01", "Q3D", lubridate::ymd("2021-01-08"), lubridate::ymd("2021-01-15"),
+#'   "P01", "EVERY 2 WEEKS", lubridate::ymd("2021-01-15"), lubridate::ymd("2021-01-29")
+#' )
+#'
+#' create_single_dose_dataset(data, by_vars = vars(USUBJID))
+create_single_dose_dataset <- function(dataset,
+                                       by_vars,
```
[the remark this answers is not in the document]
bundfussr: The `by_vars` parameter is used in the `group_by()` call only. Thus I wonder if we could remove the parameter and update the code to
```
dataset <- derive_var_obs_number(dataset, new_var = temp_obs_nr)

dataset <- dataset[rep(row.names(dataset), dataset$dose_count), ]

# Determine amount of days to adjust start_date and end_date

dataset <- dataset %>%
  group_by(temp_obs_nr) %>%
```
Then the user does not need to bother about the correct value for `by_vars`.

Or did I miss something?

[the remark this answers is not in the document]
amsmith214: i'm implicitly using the grouping to get to `row_number` for use in `time_increment`, which i want to reset from (at least) subject to subject. i don't really see how to get around not having at least a grouping by subject. if there's another approach i'm happy to pass this issue to somebody else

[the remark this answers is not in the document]
amsmith214: can you help me understand how `derive_var_obs_number` will know to increment `temp_obs_nr` appropriately without knowing the grouping? the former also seems to have a `by_vars` parameter that is set to NULL, so won't it increment `temp_obs_nr` for every record?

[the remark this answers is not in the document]
amsmith214: nvm, i see

amsmith214:
i still have to add ADEX vignette language around this function but i've got it passing everything but 3.5 (seems like period objects were not supported?)

R/create_single_dose_dataset.R
bundfussr: Please update `_pkgdown.yml` to add a "Metadata" section to the reference page.

R/create_single_dose_dataset.R
bundfussr: Could you add comment before the second example to separate them?

bms63: are the lubridate calls necessary?  short of clutters up the example

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+  assert_data_frame(lookup_table, required_vars = vars(DoseWindow, DoseCount, DayConversionFactor))
+
+    # Set up lookup table to be joined to dataset
+
+    lookup <- lookup_table %>%
+      rename(!!dose_freq := !!lookup_column)
+
+    # Use compute_duration to determine the number of completed dose periods
+
+    dataset <- dataset %>%
```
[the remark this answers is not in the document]
bundfussr: You created a nice error message but unfortunately you forgot the `abort()` call.

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,272 @@
...
+#'   A date or date-time object is expected.
+#'
+#'   Default: `AENDT`
+#'
+#' @param lookup_table The dose frequency value lookup table
+#'
+#'   The table used to look up `dose_freq` values and determine the appropriate
+#'   multiplier to be used for row generation. If a lookup table other than the
+#'   default is used, it must have columns `DoseWindow`, `DoseCount`, and
+#'   `DayConversionFactor`
```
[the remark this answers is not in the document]
bundfussr: Could you add a link to the "Pre-Defined Dose Frequencies" page or copy the "Details" section from there?

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,378 @@
...
+      derive_var_obs_number(new_var = grpseq)
+
+    # Flag to determine if date or datetime must be returned
+
+    time_flag <- nrow(dataset_part_2 %>%
+                         filter(DoseWindow %in% c("MINUTE", "HOUR"))) > 0
+
+    if (time_flag) {
+      dataset_part_1 <- dataset_part_1 %>%
+        mutate(!!start_date := as_datetime(!!start_date),
```
bundfussr, line 331: I think the `start_date` and `end_date` column should not be changed.

Somehow this seems to create issues.

For example the following example fails.
```
tex <- tibble::tribble(
    ~USUBJID, ~ASTDTM, ~AENDTM, ~EXDOSFRQ, ~EXTRT, ~temp_obs_nr,
    "1", ymd_hms("2020-02-07T02:00:00"), ymd_hms("2020-02-09T01:00:00"), "QD", "DRUG 1", 1,
    "1", ymd_hms("2020-02-07T02:00:00"), ymd_hms("2020-02-09T01:00:00"), "QD", "DRUG 2", 2
) %>%
    mutate(ASTDT = date(ASTDTM),
           AENDT = date(AENDTM))

create_single_dose_dataset(tex,
                           start_date = ASTDTM,
                           end_date = AENDTM)
```

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,378 @@
...
+
+  col_names <- colnames(dataset)
+  dose_freq <- assert_symbol(enquo(dose_freq))
+  lookup_column <- assert_symbol(enquo(lookup_column))
+  start_date <- assert_symbol(enquo(start_date))
+  end_date <- assert_symbol(enquo(end_date))
+  assert_data_frame(dataset, required_vars = quo_c(dose_freq, start_date, end_date))
+  assert_data_frame(lookup_table, required_vars = vars(DoseWindow, DoseCount, ConversionFactor))
+
+    # Set up lookup table to be joined to dataset
```
bundfussr, line 280: Could you unify the indentation?

bundfussr (changes requested):

R/create_single_dose_dataset.R
amsmith214: i think this fails because currently the `time_flag` is set based on the explicit presence of a dose frequency like HOUR or MINUTE in the data, so it's trying to turn ASTDTM/AENDTM into regular date objects.

it seems to me like the function might need to return dates for `start_date`/`end_date` in some cases and datetimes for the same variables in other cases. `time_flag` was my attempt at determining when to do this, but here were my other considerations:

- `start_time` and `end_time` in addition to `start_date` and `end_date`
- expand the conditions for when `time_flag` is positive to see if the arguments given to `start_date` and `end_date` are datetimes

this second one seems more in line with your suggestion. i think i'd still need to change the type of the column in a case like below, where a date is fed as `start_date`/`end_date` but the frequency involves hours.

```
tex <- tibble::tribble(
    ~USUBJID, ~ASTDTM, ~AENDTM, ~EXDOSFRQ, ~EXTRT, ~temp_obs_nr,
    "1", ymd_hms("2020-02-07T02:00:00"), ymd_hms("2020-02-09T01:00:00"), "Q4H", "DRUG 1", 1,
    "1", ymd_hms("2020-02-07T02:00:00"), ymd_hms("2020-02-09T01:00:00"), "Q4H", "DRUG 2", 2
) %>%
    mutate(ASTDT = date(ASTDTM),
           AENDT = date(AENDTM))

create_single_dose_dataset(tex,
                           start_date = ASTDT,
                           end_date = AENDT)
```

R/create_single_dose_dataset.R
bundfussr: You need to add `library(lubridate)` like in https://pharmaverse.github.io/admiral/reference/derive_vars_dy.html#examples.

R/create_single_dose_dataset.R
bundfussr: We should not change the type of the input variables. If the type of `ASTDT` is changed from date to datetime, it is quite confusing and not ADaM compliant.

I think we have two options:

- We expect a datetime variable for `start_date` and `end_date` and stop if another type of variable is provided.
- We accept date and datetime variables for `start_date` and `end_date`. In the function we create a temporary datetime variable for the computation and convert it to a date variable at the end (if necessary). If a date variable is provided and the computation uses hours or minutes, we could issue a warning or error.

amsmith214: sounds good. i will implement the second option. thanks for your patience and guidance!

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,384 @@
...
+  value_check <- dataset %>%
+    select(!!dose_freq) %>%
+    anti_join(lookup, by = as.character(quo_get_expr(dose_freq))) %>%
+    unique()
+
+  if (nrow(value_check) > 0) {
+    values_not_found <- str_c(value_check %>% select(!!dose_freq), collapse = ", ")
+
+    err_msg <- paste0(
+      sprintf("The following values of %s in %s do not appear in %s:\\n",
```
bundfussr, line 302: Typo: "\\n" instead of "\n"

R/create_single_dose_dataset.R
```diff
@@ -0,0 +1,384 @@
...
+
+  # Check values of lookup vs. data and return error if values are not covered
+
+  value_check <- dataset %>%
+    select(!!dose_freq) %>%
+    anti_join(lookup, by = as.character(quo_get_expr(dose_freq))) %>%
+    unique()
+
+  if (nrow(value_check) > 0) {
+    values_not_found <- str_c(value_check %>% select(!!dose_freq), collapse = ", ")
```
bundfussr, line 299: This causes a warning:
```
In addition: Warning message:
In stri_c(..., sep = sep, collapse = collapse, ignore_null = TRUE) :
  argument is not an atomic vector; coercing
```
E.g., when calling
```
tex <- tibble::tribble(
  ~USUBJID, ~ASTDTM, ~AENDTM, ~EXDOSFRQ, ~EXTRT, ~temp_obs_nr,
  "1", ymd_hms("2020-02-07T02:00:00"), ymd_hms("2020-02-09T01:00:00"), "QUID", "DRUG 1", 1,
  "1", ymd_hms("2020-02-07T02:00:00"), ymd_hms("2020-02-09T01:00:00"), "QUOD", "DRUG 2", 2
) %>%
  mutate(ASTDT = date(ASTDTM),
         AENDT = date(AENDTM))

create_single_dose_dataset(tex,
                           start_date = ASTDTM,
                           end_date = AENDTM)
```

bundfussr (changes requested):

R/create_single_dose_dataset.R
amsmith214: switched to `paste0` and seems to be fine now

Issue #578, opened 2021-10-04
Implement join function allowing for value comparisons


aehmann-gsk:
Implement function that will do something like:
`select * from data1 a, data2 b
where b.dtb <= a.dta <= b.dtb2`

See post here: https://stackoverflow.com/questions/46339638/join-tables-based-on-multiple-ranges-in-r

PR #782, opened 2021-12-20
607_merge_experiment: Implement merge functions

bundfussr:


tests/testthat/test-derive_merged.R
```diff
@@ -0,0 +1,292 @@
...
+    select(-COUNTRY)
+
+
+  expect_dfs_equal(base = expected,
+                   compare = actual,
+                   keys = c("USUBJID", "AVISIT"))
+})
+
+## derive_var_merged_cat: merge categorized variable and define value for non-matched by groups ----
+test_that("derive_vars_merged_cat: merge categorized variable and define value for non-matched by groups", {
```
konstantinakoukR, line 159: lint fails at this line, in case you missed that :)

bundfussr: Thanks, I'll fix it together with the replacement of the internal `left_join()` calls.

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#' # derive treatment start date (TRTSDT)
+#' derive_vars_merged_dt(
+#'   select(dm, STUDYID, USUBJID),
+#'   dataset_add = ex,
+#'   by_vars = vars(STUDYID, USUBJID),
+#'   new_vars_prefix = "TRTS",
+#'   dtc = EXSTDTC,
+#'   date_imputation = "first",
+#'   order = vars(TRTSDT),
```
kabis-ops, line 225: It sounds strange to me to use the derived variable as input to the function.
Should it refers only to variable in dataset_add and then inside the function, the derived variable is used in addition to the specified var?
(just thinking)
` order = vars(EXTRT, EXSEQ)` 
and when you call derive_meged_vars you add the derived variable to the order var?: 
`order = vars(EXTRT, EXSEQ, TRTSDT)`

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#' data("dm")
+#' data("vs")
+#'
+#' wgt_cat <- function(wgt) {
+#'   case_when(wgt < 50 ~ "low",
+#'             wgt > 90 ~ "high",
+#'             TRUE ~ "normal")
+#' }
+#'
+#' derive_var_merged_cat(
```
kabis-ops, line 436: Should add example with `missing_value`

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#' @param by_vars Grouping variables
+#'
+#'   *Permitted Values*: list of variables
+#'
+#' @param new_var New variable
+#'
+#'   The specified variable is added to the input dataset.
+#'
+#' @param condition Condition
```
kabis-ops, line 495: Is that needed? The `filter_add` param should do the filtering?

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+  derive_vars_merged(
+    dataset,
+    dataset_add = add_data,
+    by_vars = by_vars,
+    new_vars = vars(!!new_var),
+    order = vars(!!new_var),
+    check_type = "none",
+    mode = "last"
+  ) %>%
+    mutate(!!new_var := if_else(!!new_var == 1, true_value, false_value, false_value))
```
kabis-ops, line 568: false_value is given twice

R/derive_merged.R
```diff
@@ -0,0 +1,708 @@
...
+#'   If the uniqueness check fails, the specified message is displayed.
+#'
+#'   *Default*:
+#'
+#'   ```
+#'   paste("Dataset `dataset_add` contains duplicate records with respect to",
+#'         enumerate(vars2chr(by_vars))
+#'   ```
+#'
+#' @details
```
yangc71, line 83: Can we add some clarification this is essentially a left join (e.g. all records from the left `dataset` are kept and records from the right `dataset_add` are not included if there are no matching records in the left `dataset`)?

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'   If the parameter is not specified or set to `NULL`, all variables from the
+#'   additional dataset are added.
+#'
+#'   *Permitted Values*: list of variables created by `vars()`
+#'
+#' @param mode Selection mode
+#'
+#'   Determines if the first or last observation is selected.
+#'
+#'   *Permitted Values*: `"first"`, `"last"`
```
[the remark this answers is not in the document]
yangc71: Maybe we can clarify `mode` is required (permitted value `"first"` and `"last"`) if `order` variable is provided, else `mode` is defaulted to `NULL` if `order` is not provided?

R/derive_merged.R
```diff
@@ -0,0 +1,708 @@
...
+      date_imputation = date_imputation,
+      flag_imputation = flag_imputation,
+      min_dates = min_dates,
+      max_dates = max_dates
+    )
+  new_vars <- quos(!!!syms(setdiff(names(add_data), old_vars)))
+  derive_vars_merged(dataset,
+                     dataset_add = add_data,
+                     by_vars = by_vars,
+                     order = order,
```
yangc71, line 286: When `date_imputation` is NULL and a partial date exists, then the `derive_vars_dt` step converts the partial dates to missing. As a result, the date variable specified in `order` dictates which record is chosen. For example, if there are two records with `EXSTDTC` being`“2020”` and `“2020-12-14”`, then converted `TRTSDT` are `NA` and `2020-12-14`. Assume `mode = “first”`, if `order = vars(TRTSDT)` (converted date) then the record with complete date is chosen. But if `order = vars(EXSTDTC)` (unconverted character date), then the record with partial date is chosen. I’m not sure if this is intended or I'm making this too confusing. If it's intended, should we include some clarifications about which records are chosen when partial dates are not imputed?

R/derive_merged.R
```diff
@@ -0,0 +1,708 @@
...
+                                        dataset_add,
+                                        by_vars,
+                                        order = NULL,
+                                        new_var,
+                                        source_var,
+                                        case = NULL,
+                                        filter_add = NULL,
+                                        mode = NULL,
+                                        missing_value = NA_character_) {
+
```
yangc71, line 671: I understand we created multiple merging functions to avoid users writing expressions in input. But in general for these wrapper functions, I'm wondering is it useful to allow users to keep more source variables (without transformation)? So they don't have to run merging multiple times if they want to merge in multiple variables (for example, from my understanding if they want to merge a converted character variable, but in addition also keep other variables as-is from `dataset_add` they have to run `derive_var_merged_character` then `derive_vars_merged`).

yangc71 (commented):
Hi Stefan, apologies for the delayed review, I was a bit busy with study work this month. I have added my comments, and I know I missed several meetings, so please ignore any comments that have already been discussed or are not applicable.

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
+#' Merge Variables from a Dataset to the Input Dataset
+#'
+#' Merge variables from a dataset to the input dataset. The observations to
+#' merge can be selected by a condition and/or selecting the first or last
```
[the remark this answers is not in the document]
bundfussr: Yes, updated

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#' observation for each by group.
+#'
+#' @param dataset Input dataset
+#'
+#'   The variables specified by the `by_vars` parameter are expected.
+#'
+#' @param dataset_add Additional dataset
+#'
+#'   The variables specified by the `by_vars`, the `new_vars`, and the `order`
```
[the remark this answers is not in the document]
bundfussr: Added

R/derive_merged.R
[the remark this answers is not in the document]
bundfussr: Updated

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'   parameter are expected.
+#'
+#' @param order Sort order
+#'
+#'   The first or last observation from the addtional dataset is selected with
+#'   respect to the specified order.
+#'
+#'   *Permitted Values*: list of variables or `desc(<variable>)` function calls
```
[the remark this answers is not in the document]
bundfussr: Updated

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#'   *Permitted Values*: `"first"`, `"last"`
+#'
+#' @param by_vars Grouping variables
+#'
+#'   The input dataset and the selected observations from the additional dataset
+#'   are merged by the specified by variables. The by variables must be a unique
+#'   key of the selected observations.
+#'
+#'   *Permitted Values*: list of variables
```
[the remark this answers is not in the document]
bundfussr: Updated

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'   with respect to the by variables and the order.
+#'
+#'   *Permitted Values*: `"none"`, `"warning"`, `"error"`
+#'
+#' @details
+#'
+#'   1. The additional dataset is restricted to the observations matching the
+#'   `filter_add` condition.
+#'
+#'   1. If `order` is specified, for each by group the first or last observation
```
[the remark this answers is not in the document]
bundfussr: `1.` is just a placeholder to indicate a numbered list. In the rendered version it is replaced by the correct number.

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#' # merge last weight to adsl
+#' data("adsl")
+#' derive_vars_merged(
+#'   adsl,
+#'   dataset_add = vs,
+#'   by_vars = vars(STUDYID, USUBJID),
+#'   order = vars(VSDTC),
+#'   mode = "last",
+#'   new_vars = vars(LASTWGT = VSSTRESN, LASTWGTU = VSSTRESU),
+#'   filter_add = VSTESTCD == "WEIGHT"
```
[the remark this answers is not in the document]
bundfussr: For the merge functions the selection is based on `dataset_add` only.

I agree that we should have a function for your use case. We have already an issue for it (https://github.com/pharmaverse/admiral/issues/578).

I would not integrate this functionality into the merge functions because the flow is different and it requires much more resources. The datasets need to be joined first and then restricted. Joining first can lead to huge datasets. Thus it should be done only if it is really required.

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'   *Permitted Values*: list of variables or `desc(<variable>)` function calls
+#'
+#' @param new_vars Variables to add
+#'
+#'   The specified variables are added to the output dataset. Variables can be
+#'   renamed by naming the element, i.e., `new_vars = vars(<new name> = <old
+#'   name>)`.
+#'
+#'   If the parameter is not specified or set to `NULL`, all variables from the
+#'   additional dataset are added.
```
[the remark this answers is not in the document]
bundfussr: Added

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
+#' Merge Variables from a Dataset to the Input Dataset
```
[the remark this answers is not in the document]
bundfussr: Updated

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#'   The first or last observation from the addtional dataset is selected with
+#'   respect to the specified order.
+#'
+#'   *Permitted Values*: list of variables or `desc(<variable>)` function calls
+#'
+#' @param new_vars Variables to add
+#'
+#'   The specified variables are added to the output dataset. Variables can be
+#'   renamed by naming the element, i.e., `new_vars = vars(<new name> = <old
```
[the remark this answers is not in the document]
bundfussr: I have updated the description. Is it clear now?

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#' @param by_vars Grouping variables
+#'
+#'   The input dataset and the selected observations from the additional dataset
+#'   are merged by the specified by variables. The by variables must be a unique
+#'   key of the selected observations.
+#'
+#'   *Permitted Values*: list of variables
+#'
+#' @param filter_add Filter for additional data
```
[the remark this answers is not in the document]
bundfussr: Added

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#' @details
+#'
+#'   1. The additional dataset is restricted to the observations matching the
+#'   `filter_add` condition.
+#'
+#'   1. If `order` is specified, for each by group the first or last observation
+#'   (depending on `mode`) is selected.
+#'
+#'   1. The variables specified for `new_vars` are merged to the input dataset.
```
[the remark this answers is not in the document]
bundfussr: I am not sure. "created" sounds a bit confusing to me. What about "are renamed (if requested) and merged"?

R/derive_merged.R
```diff
@@ -0,0 +1,685 @@
...
+#'
+#'   If `"warning"` or `"error"` is specified, the specified message is issued
+#'   if the observations of the (restricted) additional dataset are not unique
+#'   with respect to the by variables and the order.
+#'
+#'   *Permitted Values*: `"none"`, `"warning"`, `"error"`
+#'
+#' @details
+#'
+#'   1. The additional dataset is restricted to the observations matching the
```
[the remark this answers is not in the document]
bundfussr: Updated

R/derive_merged.R
bundfussr: Good idea. Done

R/derive_merged.R
bundfussr: In principle yes, but I wanted to distinguish between no data and condition not fulfilled.

I have added an example (and the `missing_value` parameter).

R/derive_merged.R
bundfussr: This is intended the first one is for condition is false and the second one for condition is `NA`.

However, I have replaced the second one by `missing_value`.

R/derive_merged.R
```diff
@@ -0,0 +1,708 @@
...
+#'   The variables specified by the `by_vars` parameter are expected.
+#'
+#' @param dataset_add Additional dataset
+#'
+#'   The variables specified by the `by_vars`, the `new_vars`, and the `order`
+#'   parameter are expected.
+#'
+#' @param order Sort order
+#'
+#'   The first or last observation from the addtional dataset is selected with
```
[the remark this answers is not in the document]
bundfussr: Fixed

R/derive_merged.R
```diff
@@ -0,0 +1,708 @@
...
+                               filter_add = NULL,
+                               match_flag = NULL,
+                               check_type = "warning",
+                               duplicate_msg = NULL) {
+  filter_add <- assert_filter_cond(enquo(filter_add), optional = TRUE)
+  assert_vars(by_vars)
+  assert_order_vars(order, optional = TRUE)
+  assert_vars(new_vars, optional = TRUE)
+  assert_data_frame(dataset, required_vars = by_vars)
+  assert_data_frame(dataset_add, required_vars = quo_c(by_vars, extract_vars(order)))
```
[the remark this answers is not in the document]
bundfussr: Good catch! `new_vars` added.

R/derive_merged.R
bundfussr: I have added more details. Is it clear now?

R/derive_merged.R
bundfussr: Good point! I have added some details to the description of the `order` parameter. I would expect that in most cases missings are excluded by `filter_add`.

However, the current behaviour looks a bit inconsistent. For `mode = "first"` observations with `NA` are selected only if all observation are `NA` while for `mode = "last"` observations with `NA` are always selected. Maybe we should change this behaviour. What do you think?

R/derive_merged.R
bundfussr: You are right, with our current approach the user may need to call `derive_vars_merged_*` functions more than once for merging multiple variables. This was not my favourite solution but the majority preferred this approach. We have discussed it multiple times. Thus I would not like to discuss it again. At least not now. I would wait for user feedback before rediscussing it.

bundfussr:
Hi @kabis-ops , @yangc71 , thanks for your review and all your comments. I hope I have addressed all of them.

R/derive_merged.R
yangc71: Yes, thanks!

R/derive_merged.R
yangc71: Thanks for adding the details in the description. 

If I'm understanding correctly, this behavior seems to be carried from `filter_extreme`. I think in some cases the users may not want to choose records with `NA`. Maybe we can add a parameter for `filter_extreme` (e.g. `remove_na`) to allow users to choose whether `NA` should be considered?

R/derive_merged.R
yangc71: I see. Thanks for the information.

R/derive_merged.R
bundfussr: Yes, you are correct. I have added it to the agenda of the next "Questions and Comments" meeting as it affects many functions.

bundfussr: Hi @yangc71 , we discussed the topic today and decided to keep the behavior as it is because it is in line with the `arrange()` function. We also discussed to provide a `remove_na` parameter but decided not to provide it because usually `filter_extreme` is not called directly but within a derivation and the derivations already provide a `filter` parameter to exclude NAs.

R/derive_merged.R
```diff
@@ -0,0 +1,848 @@
...
+#' @param order Sort order
+#'
+#'   If the parameter is set to a non-null value, for each by group the first or
+#'   last observation from the additional dataset is selected with respect to the
+#'   specified order.
+#'
+#'   *Default*: `NULL`
+#'
+#'   *Permitted Values*: list of variables or `desc(<variable>)` function calls
+#'   created by `vars()`, e.g., `vars(ADT, desc(AVAL)` or `NULL`
```
kabis-ops, line 27: there is a missing parethesis after AVAL)

R/derive_merged.R
kabis-ops: OK, it may be good to add some details for the `condition` parameter, then?

R/derive_merged.R
bundfussr: Thanks, fixed

R/derive_merged.R
bundfussr: Yes, I have added a description. I have also updated the details section as the `missing_value` parameter was not mentioned there.

yangc71 (approved):

PR #783, opened 2021-12-21
678 Update faq.Rmd

rossfarrugia:
678 added detail on validation to FAQ - @koegerr i don't think we need to add much more than the extra bullet i added here for now. James is still involved with wider discussions on the GitHub action with GSK and another company. the hope is this will become a recommended way for the industry to validate r packages.


rossfarrugia:
@bundfussr are you ok to review and merge this small update please?

bundfussr (approved):

Issue #744, opened 2021-12-08
notin function not being loaded if adsl template is followed


teckla-gsk:
In format_dcsreas  section of adsl template notin was not being loaded as part of admiral??? Something to investigate

thomas-neitmann:
Could you elaborate? I don't quite get it...

rossfarrugia:
@teckla-gsk @thomas-neitmann i was just working on another ADSL template issue so i looked into this. it's not actually used in the template but it is referenced in a section of the ADSL vignette here: https://roche-gsk.github.io/admiral/articles/adsl.html#disposition_reason, where it talks about "Example function format_dcsreas():"

the function "notin" is covered here https://roche-gsk.github.io/admiral/reference/grapes-notin-grapes.html and is covered by our utils.R code. i tried running the ADSL vignette code extract and it worked for me.

i think ok to close this issue, but feel free to re-open if you think i missed anything.

thomas-neitmann:
We probably should not use `%notin%` in the `format_dcsreas()` ` function. The former is only meant for internal purposes but the latter s exposed to users which could lead to confusion. Instead of `a %notin% b` we can just use `!a %in% b`.

rossfarrugia:
thanks @thomas-neitmann - now updated

PR #784, opened 2021-12-21
749-ADSL template and vignette updates to use agegr function

rossfarrugia:
I updated the ADSL template to use derive_agegr_fda rather than a user defined function. When I was doing this i also spotted the LSTALVDT code was falling over as we didn't read in AE and LB test datasets at the start so i added these, Also updated the ADSL vignette to include the agegr FDA example but also keep a separate example for user-defined age group, in case users ever need to do this.
Note I didn't update the NEWS.md file as i didn't deem this update worthy as it's so minor, but shout if you disagree.

### DIFF
--- a/inst/templates/ad_adsl.R
+++ b/inst/templates/ad_adsl.R
@@ -17,10 +17,14 @@ library(lubridate)
 data("dm")
 data("ds")
 data("ex")
+data("ae")
+data("lb")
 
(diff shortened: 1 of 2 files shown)


rossfarrugia:
@millerg23 FYI made minor update for #744 as per Thomas's comment whilst i was updating ADSL template with this PR anyway

thomas-neitmann (approved):

PR #804, opened 2022-01-06
Update adsl.Rmd

statasaurus:
Update to more standard code. It is more explicit to select(- col) rather than just mutate and set it to NULL


thomas-neitmann:
@statasaurus Could you please open a new PR and make sure that you

1. create your new branch of `devel` not `master`
2. the PR is directed at `devel` not `master` and
3. your branch name contains underscores not dashes

See also our [git and GitHub Guidance](https://roche-gsk.github.io/admiral/articles/git_usage.html).

Thanks!

PR #924, opened 2022-02-12
Test GH package installation for validation

dinakar29:
Given that `admiral.test` is in the `pharmaverse` org, simply use the `Remotes` field in the DESCRIPTION to install so that the validation reports pass.

### DIFF
--- a/.github/workflows/r-pkg-validation.yml
+++ b/.github/workflows/r-pkg-validation.yml
@@ -3,13 +3,16 @@ name: R Package Validation report
 on: # Run this action when a release is published
   release:
     types: [published]
+  pull_request_target:
+    branches:
+      - main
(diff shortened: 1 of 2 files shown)


thomas-neitmann:
Thanks @dinakar29! I approved the GH actions runs. Please note that we submitted {admiral.test} to CRAN (along with {admiral} itself) which is why I removed the `Remotes` field which CRAN does not allow.

dinakar29:
> Thanks @dinakar29! I approved the GH actions runs. Please note that we submitted {admiral.test} to CRAN (along with {admiral} itself) which is why I removed the `Remotes` field which CRAN does not allow.

Noted. Thanks, I'll remove that field altogether.

PR #1135, opened 2022-05-14
Fix the Roxygen workflow

dinakar29:
- [x] Fix the Roxygen workflow
- [x] Use plain markdown code highlighting for unevaluated vignette chunk so the `styler` does not need to reformat it

### DIFF
--- a/vignettes/programming_strategy.Rmd
+++ b/vignettes/programming_strategy.Rmd
@@ -426,7 +426,7 @@ The documentation will be updated at:
 
 + the description level for a function, 
 
-    ```r
+    ```{r, eval=FALSE}
     #' Title of the function
     #'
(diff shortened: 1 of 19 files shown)


vignettes/programming_strategy.Rmd
```diff
@@ -426,7 +426,7 @@ The documentation will be updated at:
 
 + the description level for a function, 
 
-    ```{r, eval=FALSE}
+    ```r
```
bundfussr, line 429: Could you undo this change? It is intended that the R code is not executed. This allows developers to copy the code to the function they want to deprecate.

bundfussr (changes requested):

bundfussr (approved):

PR #1161, opened 2022-05-20
Fix man pages w/f + more badeges!

dinakar29:
- [x] Fix the man pages workflow
- [x] Add more badges to README

### DIFF
--- a/.github/workflows/man-pages.yml
+++ b/.github/workflows/man-pages.yml
@@ -103,6 +103,7 @@ jobs:
           git commit -m "Update man pages"
 
       - name: Push updates ⤴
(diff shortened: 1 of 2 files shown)


bms63 (approved):
Cool thanks.  

Also, where are the secrets stored.  I store stuff like that in a .file in my home environment for databases and such but since this is an action how does one update secrets.

dinakar29:
Github offers a way to store encrypted secrets, and it also produces dynamic secrets for workflows.

bms63:
Also @dinakar29 are you on slack?  what is your name :)

PR #1175, opened 2022-05-25
Closes #1156 - Code coverage reports without codecov

### LINKED ISSUE #1156: Set up code coverage badge for pharmaverse/admiral
Pointing to old repo

dinakar29:
Replace codecov with custom workflow.


dinakar29:
Detailed coverage reports are uploaded to the action logs and can be downloaded there.
Screen Shot 2022-05-25 at 2 15 00 PM

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,185 @@
+on:
+  workflow_dispatch:
+  push:
+    branches:
+      - main
+      - master
```
bms63, line 6: should we just remove master since we don't use that branch name anymore?

.github/workflows/code-coverage.yml
```diff
@@ -0,0 +1,185 @@
...
+        run: |
+          tryCatch(
+            expr = {
+                x <- covr::package_coverage(
+                  path=".",
+                  clean = FALSE,
+                  quiet = FALSE
+                )
+                print(x)
+                covr::to_cobertura(x, filename = "coverage.xml")
```
bms63, line 76: I was wondering how to convert the report to xml.  this is really cool!

.github/workflows/code-coverage.yml
dinakar29: Yeah we can. I'll remove it wherever it's being referenced.

.github/workflows/code-coverage.yml
dinakar29: Yeah this is definitely one of the nice things about the package. We can export to different formats.

bms63 (approved):
Looks really nice!  This isn't rendering on branch
image

dinakar29:
> Looks really nice! This isn't rendering on branch image

Rendering will take place after this has been merged to `devel`, since it's pointing to that branch.
If you look at the `badges` branch, each directory will contain the badge for the corresponding branch, so when the workflow runs for `devel`, the badge will be added there.

bms63:
thanks for being so thorough on .yml files.

bms63 (approved):

rossfarrugia:
thanks @dinakar29 - looks awesome!

PR #1181, opened 2022-05-27
Let coverage badge show %

dinakar29:
Instead of just the number.


.github/workflows/code-coverage.yml
```diff
@@ -151,7 +151,7 @@ jobs:
       id: badge
       with:
         label: 'Coverage'
```
bms63, line 153:
```suggestion
        label: 'Test Coverage'
```

bms63 (approved):
Whoa...so fast

PR #1237, opened 2022-06-23
Change linter to superlinter

dinakar29:
Given that it's the best one on the market.


dinakar29:
Closing this, turns out superlinter doesn't play well with older R versions.

PR #1254, opened 2022-06-28
Revert "Add multi-version docs job"

dinakar29:
Reverts pharmaverse/admiral#1245

### DIFF
--- a/.github/workflows/pkgdown.yml
+++ b/.github/workflows/pkgdown.yml
@@ -1,11 +1,8 @@
 on:
   workflow_dispatch:
-  workflow_call:
   push:
     branches:
       - main
-      - devel
-      - pre-release
 
 name: pkgdown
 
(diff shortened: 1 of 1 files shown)


bundfussr (approved):

bms63 (approved):
Thanks!  I like the idea of using admiraltemplate and admiralonco to test this out and get it right.

bms63:
Umm...did something get updated in devel that is causing the link checker to fail now?  I thought none of the PRs made it in

thomas-neitmann:
Everything fails with "Network error" so I assume it's not related to the links themselves.

bms63:
This needs to go to `main` and not `devel` so the website republishes

dinakar29:
This just needs to go into `devel`, I'll clean up the `gh-pages` branch after.

PR #1274, opened 2022-06-30
adding google analytics

epijim:
as discussed with @thomas-neitmann and @bms63

this only adds some HTML to the pkgdown site - it has no impact on the R package itself

### DIFF
--- a/_pkgdown.yml
+++ b/_pkgdown.yml
@@ -3,7 +3,18 @@ url: https://pharmaverse.github.io/admiral
 template:
   bootstrap: 5
   params:
-    bootswatch: flatly
+    bootswatch: flatly'
+  includes:
+    in_header: |
+       <!-- Global site tag (gtag.js) - Google Analytics -->
+       <script async src="https://www.googletagmanager.com/gtag/js?id=G-574L344W3R"#' ></script>
+       <script>
+         window.dataLayer = window.dataLayer || [];
+         function gtag(){dataLayer.push(arguments);}
+         gtag('js', new Date());
+
+         gtag('config', '{YOUR TRACKING ID}');
+       </script>
 search:
   exclude: ['news/index.html']
 repo:


epijim:
hmmm - not 100% sure I got the indent right here

epijim:
Ok - looks right in the file. The '+' makes it look wonky in the PR viewer.

image

bms63 (approved):

bms63:
I just realized that this should of gone to `main` as that is where the website is published to.

bms63:
Also we should branch off main for this PR as we don't want all the dev branch stuff in main.  Is this ` okay after flatly?

```
bootswatch: flatly'
  includes:
    in_header: |
       <!-- Global site tag (gtag.js) - Google Analytics -->
       <script async src="https://www.googletagmanager.com/gtag/js?id=G-574L344W3R"#' ></script>
       <script>
         window.dataLayer = window.dataLayer || [];
         function gtag(){dataLayer.push(arguments);}
         gtag('js', new Date());
         gtag('config', '{YOUR TRACKING ID}');
       </script>
```

thomas-neitmann:
@epijim @bms63 I'm actually very much **against** using Google Analytics. Their tracking is super intrusive and it's a shame that you're literally being followed around by Google anywhere you go on the web. If we want to have some kind of analytics for the website I suggest we use a privacy respecting alternative such as [Plausible](https://plausible.io/privacy-focused-web-analytics).

epijim:
We'd need to pay to use plausible, as they sell software (to your point, with them we are not the product). 


While I agree with the general sentiment - I wonder if it's a little misplaced when talking about a site primarily used in a corporate context (eg pharma employees). Far more is tracked on the device you are using. 

If we were talking about a site focussed on people's non-work related activity I think I would be much more behind the sentiment.

epijim:
(Note we can pay - it's just quite  informal at the moment. Pharmaverse AWS activity is costing < 1CHF a month, so it's going against my card at the moment. This would make our monthly costs something like 11 CHF.

PR #1302, opened 2022-07-08
Closes #1293 Implement derive_var_confirmation_flag()

### LINKED ISSUE #1293: Feature Request: Implement `derive_var_confirmation_flag()`
### Feature Idea

Implementing the following function.
```
derive_var_confirmation_flag(
  dataset,
  by_vars,
  order,
  new_var,
  join_vars,
  first_cond = NULL,
  filter,
  check_type = "warning"
)
```

The parameters are the same as for `filter_confirmation()` (see https://github.com/pharmaverse/admiral/issues/1292) except the additional parameter `new_var`.
The function calls `filter_confirmation()` and then flaga all observations of the input dataset which are returned by the `filter_confirmation()` call.

### Relevant Input

_No response_

### Relevant Output

_No response_

### Reproducible Example/Pseudo Code

```
derive_var_confirmation_flag(
    adae,
    by_vars = vars(STUDYID, USUBJID),
    order = vars(ASTDTM, AESEQ),
    new_var = ALCOVFL,
    join_vars = vars(ASTDTM, ACCOVFL),
    filter = date(ASTDTM) >= date(ASTDTM.join) + days 7 & ACCOVFL.join == "Y" &
      (ADURN > 30 | AEOUT == "NOT RECOVERED/NOT RESOLVED"),
  )
```
This call flags AEs which occurred at, after or up to seven days before a COVID AE (`ACCOVFL == "Y"`) and which last more than 30 days or were not resolved.

bundfussr:
Thank you for your Pull Request! We have developed this task checklist from the [Development Process Guide](https://pharmaverse.github.io/admiral/articles/development_process.html) to help with the final steps of the process. Completing the below tasks helps to ensure our reviewers can maximize their time on your code as well as making sure the admiral codebase remains robust and consistent.   

- [x] Code is formatted according to the [tidyverse style guide](https://style.tidyverse.org/) 
- [x] Updated relevant unit tests or have written new unit tests - See [Unit Test Guide](https://pharmaverse.github.io/admiral/articles/unit_test_guidance.html#writing-unit-tests-in-admiral-)
- [x] If you removed/replaced any function and/or function parameters, did you fully follow the [deprecation guidance](https://pharmaverse.github.io/admiral/articles/programming_strategy.html#deprecation-1)?
- [x] Update to all relevant roxygen headers and examples 
- [x] Update `NEWS.md` if the changes pertain to a user-facing function (i.e. it has an `@export` tag) or documentation aimed at users (rather than developers)


R/derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,329 @@
...
+#'   The variables needed from the subsequent observations should be specified
+#'   for this parameter. The specified variables are added to the joined dataset
+#'   with suffix ".join". For example to flag all observations with `AVALC ==
+#'   "Y"` and `AVALC == "Y"` for at least one subsequent visit `join_vars =
+#'   vars(AVALC, AVISITN)` and `filter = AVALC == "Y" & AVALC.join == "Y" &
+#'   AVISITN < AVISITN.join` could be specified.
+#'
+#'   The `*.join` variables are not included in the output dataset.
+#'
+#' @param join_type Observations to keep after joining
```
rossfarrugia, line 35: is this a new argument since the admiralonco version? i didn't notice it before or in the issue. i find it a bit confusing so think more is needed here in the description to explain why needed.

my understanding is when we did this for onco BOR we only ever needed to look for the records on or after first PR/CR here, but now for the long COVID example we'd actually want the option to look even before the AE of interest.

R/derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,329 @@
...
+#'   "4",      2,        "N",
+#' )
+#'
+#' derive_var_confirmation_flag(
+#'   data,
+#'   by_vars = vars(USUBJID),
+#'   new_var = CONFFL,
+#'   join_vars = vars(AVALC, AVISITN),
+#'   order = vars(AVISITN),
+#'   filter = AVALC == "Y" & AVALC.join == "Y" & AVISITN < AVISITN.join
```
rossfarrugia, line 216: looking through the examples again and trying to put myself in a neutral user perspective and i think the piece users will most struggle with understanding here is what's the difference when using `first_cond`, e.g. why would the above be different to:
```
first_cond = AVALC.join == "Y",
filter = AVALC == "Y" & AVISITN < AVISITN.join
```
Any way you think we could help explain this more or maybe via an example? i'm even struggling myself today to see the difference, but maybe it's just because it's first thing Monday morning...

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,183 @@
...
+  "3",      1,        "SD",
+  "4",      1,        "PR",
+  "4",      2,        "PD",
+  "4",      3,        "SD",
+  "4",      4,        "SD",
+  "4",      5,        "PR"
+)
+
+# derive_var_confirmation_flag ----
+## Test 1: filter without first_cond ----
```
rossfarrugia, line 21: might help to add an extra line of more detail of the derivation you are intending in these test comment explanations, e.g. Flagging any patient PR value that is followed subsequently by a CR or PR

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,183 @@
+library(tibble)
+data <- tribble(
+  ~USUBJID, ~AVISITN, ~AVALC,
+  "1",      1,        "PR",
```
rossfarrugia, line 4: could you add a patient where we can see impact of `join_type = "at_after"`, e.g. a patient who has a CR PR PR and then the CR gets ignored, and only the first PR gets flagged

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,183 @@
...
+
+  expect_dfs_equal(
+    base = expected,
+    compare = actual,
+    keys = c("USUBJID", "AVISITN")
+  )
+})
+
+## Test 4: join_type = "all" ----
+test_that("derive_var_confirmation_flag, Test 4: join_type = 'all'", {
```
rossfarrugia, line 139: would be good to see a unit test for each of the other `join_type` options too

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,183 @@
...
+  adae <- tribble(
+    ~USUBJID, ~ADY, ~ACOVFL, ~ADURN,
+    "1",        10, "N",          1,
+    "1",        21, "N",         50,
+    "1",        23, "Y",         14,
+    "1",        32, "N",         31,
```
rossfarrugia, line 145: include a >30 duration AE 8 days before the covid AE to show that this wont be flagged

R/derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,329 @@
+#' Derive Confirmation Flag
+#'
+#' Derive a flag which depends on subsequent observations of the dataset. For
+#' example, flagging events which need to be confirmed by a second event.
```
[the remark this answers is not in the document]
bundfussr: Yes, good idea.

R/derive_var_confirmation_flag.R
bundfussr: Yes, this is a new argument to control which confirmation assessments are considered. I have added some more explanation. Is it clearer now?

R/derive_var_confirmation_flag.R
bundfussr: Good point! There is no difference. Both examples work. You need `first_cond` if `filter` uses summary functions which should be applied up to the confirmation assessment only, e.g., number of NEs between assessment and confirmation assessment should be at most one.

I have added this to the documentation of `first_cond`.

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: I have updated the example (`join_type` and `filter`). Patient 4 is now showing the impact of `join_type`.

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: Added

R/derive_var_confirmation_flag.R
rossfarrugia: Minor suggested changes to your addition:

#'   For example for confirmed response or BOR in the oncology setting or
#'   confirmed deterioration in questionnaires the confirm**atory** assessment must be
#'   after the assessment to be flagged. Thus `join_type = "after"` could be
#'   used.
#'
**#'   Whereas, sometimes you might allow for confirmatory observations to 
#'   occur prior to the observation to be flagged. For example,
#'   to flag AEs occurring on or after seven days before a COVID AE.
#'   Thus `join_type = "all"` could be used.**

tests/testthat/test-derive_var_confirmation_flag.R
rossfarrugia: do same for newly added Test 5 & 6

tests/testthat/test-derive_var_confirmation_flag.R
rossfarrugia: what about adding the following to really exemplify this:
```
    "5",      1,        "PR",   NA_character_,
    "5",      1,        "CR",   NA_character_
```
As it shows you need to be after (in the unlikely event you had 2 responses at same time - but in reality they'd use date as well here instead of seq only anyway).

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: Done

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: If I add this patient, the order is not well defined and the function issues a warning. Thus I would not add it.

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: I think test 5 and 6 are not well designed. I have updated them. Is it clearer now?

rossfarrugia: i think test 6 is incorrect. i don't agree records 1 or 4 should have confirmation flagged, as it seems like the same single record is confirming itself. i always thought you need a separate observation to confirm the original observation.

tests/testthat/test-derive_var_confirmation_flag.R
rossfarrugia: ah ok, i wonder if the `at_before` and `at_after` options are even ever needed then? when would you have the scenario to use these if it fires a warning anyway if you had 2 records with matching order variables? might be simpler just to have `after`, `before` and `all` options

bundfussr: The warning is not caused by `at_before` or `at_after`. You always get it if the variables specified for `by_vars` and `order` are not a unique key. In this case the selections specified by `join_type` and `first_cond` are questionable. The user can get rid of the warning by updating `order`, `by_vars`, or `check_type`. It depends on the data and the filter conditions which of the arguments should be updated.
I think in most cases `after`, `before`, or `all` are used but sometimes `at_before` or `at_after` may be useful as well (like in test 6).

R/derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,350 @@
+#' Derive Confirmation Flag
+#'
+#' Derive a flag which depends on subsequent observations of the dataset. For
```
rossfarrugia, line 3: change "subsequent" to "other" as they could use `before` option

R/derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,350 @@
...
+#'
+#'   The observations are ordered by the specified order.
+#'
+#' @param new_var New variable
+#'
+#'   The specified variable is added to the input dataset.
+#'
+#' @param join_vars Variables to keep from joined dataset
+#'
+#'   The variables needed from the subsequent observations should be specified
```
rossfarrugia, line 33: "subsequent" again - check and update any places this is used in this documentation or `filter_confirmation`

tests/testthat/test-derive_var_confirmation_flag.R
rossfarrugia: i disagree with this - we're trying to do too many things here. the explanation of the function reads "Derive a flag which depends on **subsequent observations** of the dataset." 

i think we should remove the `at_*` options and the same record shouldn't be able to confirm itself. i don't see that as a realistic use case of this function and it just over-complicates an already complicated function for the users.

tests/testthat/test-derive_var_confirmation_flag.R
rossfarrugia: see my other comment - i'm strongly in favour of removing the `at_*` options. if you disagree we could add to the Q&A meet Weds and see what others feel.

rossfarrugia (changes requested):

R/derive_var_confirmation_flag.R
bundfussr: Good point! Updated

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: I agree that the `at_*` options will be used only in rare cases. Nevertheless I would keep them because removing them does not simplify the code, documentation, or usage and in my experience it is just a matter of time until they will be requested.

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: See my answer to the other comment.

R/derive_var_confirmation_flag.R
millerg23: I can see the benefit of having 2 input parameters ie. `filter` and `first_cond`, however I agree with @rossfarrugia it is quite confusing. Could the input parameters be renamed, so `filter` would always be used, and would not contain any summary functions (ie what `first_cond` is at the moment), and when summary functions are required you would use the second input parameter, called something like `filter_summary`?

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,269 @@
...
+
+  expect_dfs_equal(
+    base = expected,
+    compare = actual,
+    keys = c("USUBJID", "AVISITN")
+  )
+})
+
+## Test 2: filter with first_cond ----
+## Flagging any patient CR value that is followed subsequently by a CR
```
millerg23, line 61: Same comment about use of `subsequently` as test1, and should `join_type = "after"` be set,  as default is "at_after" which isn't correct for this test.

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,269 @@
...
+  )
+
+  expect_dfs_equal(
+    base = expected,
+    compare = actual,
+    keys = c("USUBJID", "AVISITN")
+  )
+})
+
+## Test 3: filter with first_cond and summary function ----
```
millerg23, line 100: Same comment for Test 3 as already made for Test 2: Use of `subsequently` and setting `join_type = "after"`. Adding the extra test requested for test 2 not necessary for Test 3, unless you really want to!

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,269 @@
...
+  )
+
+  expect_dfs_equal(
+    base = expected,
+    compare = actual,
+    keys = c("USUBJID", "AVISITN")
+  )
+})
+
+## Test 4: join_type = "all" ----
```
millerg23, line 142: I don't understand the description of Test 4. 
Code appears to be flagging where AE has duration longer than 30 days and start date >= (start date - 7 days of records with ACOVFL == "Y").

tests/testthat/test-derive_var_confirmation_flag.R
millerg23: I can see benefit of having `at_*` options, but happy to discuss further at next week meeting

R/derive_var_confirmation_flag.R
bundfussr: I think this could be misleading. I would not rename `filter` because this should be the main condition which defines whether an observation is flagged or not. I would rather rename `first_cond` because it seems that the purpose of the parameter is not clear. The purpose is to restrict the other observations for which `filter` is evaluated. (As an observation is flagged if `filter` is `TRUE` for any observation, the restriction affects the result only if summary functions are used in `filter`.)
Maybe we should rename `first_cond` to `max_obs_cond` or `last_obs_cond`?

What do you think?

bundfussr: I see. I think test 2 is not a good example. For flagging CR assessments followed by a CR assessment I would rather use
```
derive_var_confirmation_flag(
      data,
      new_var = CONFFL,
      by_vars = vars(USUBJID),
      join_vars = vars(AVALC),
      join_type = "after",
      order = vars(AVISITN),
      filter = AVALC == "CR" & AVALC.join == "CR"
    )
```
I think this is clearer than the code in test 2. (We can not avoid that the code in test 2 works.)

Maybe we should remove test 2? The current code is testing the same as test 3 and the code above would test the same as test 1.

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,269 @@
...
+      data,
+      new_var = CONFFL,
+      by_vars = vars(USUBJID),
+      join_vars = vars(AVALC),
+      first_cond = AVALC == "CR" &
+        AVALC.join == "CR",
+      order = vars(AVISITN),
+      filter = TRUE
+    )
+
```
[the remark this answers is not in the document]
bundfussr: Added

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: Removed "subsequently"

tests/testthat/test-derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,269 @@
...
+  "4",      1,        "PR",
+  "4",      2,        "PD",
+  "4",      3,        "SD",
+  "4",      4,        "SD",
+  "4",      5,        "PR"
+)
+
+# derive_var_confirmation_flag ----
+## Test 1: filter without first_cond ----
+## Flagging any patient PR value that is followed subsequently by a CR or PR
```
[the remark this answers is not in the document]
bundfussr: Removed "subsequently"

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: Removed "subsequently"

`join_type = after` was used implicitly when `first_cond` was used. As this behaviour is confusing, I have removed the implicit conditions and set the default for `join_type` to `"after"` (this is the more common use case).

tests/testthat/test-derive_var_confirmation_flag.R
koegerr: Not sure if I understand it 100% correct, so please feel free to correct me if my assessment is based on a wrong assumption.

at_* seems to function as a simple flag based on an occurrence of a single AE, while the purpose of the function is to confirm an AE due to having another AE present.: 

 _Derive a flag which depends on other observations of the dataset. For
#' example, flagging events which need to be confirmed by a second event._

I would keep the functionality down to the purpose of "confirming by a second event", to keep it in line with our manifesto in terms of simplicity and not overload functions with too many options. The function itself seems already be complex in terms of parameters, by adding at_* I feel we actually changing the original meaning of the function itself. 
Here I would really like to keep the function clean to have one purpose, which is confirmation by a second event.

Personally (if I understood the at_* purpose correctly), if I only would like to flag a AE depending on itself, let's say Special AE or a specific AE + Severity or Hospitalization, I would just do this with one line of conditional code: it is simple to read and transparent. 
For example, If I would take over some code and see this function called with the at_* parameter, I first to understand the function and its option and I would always ask myself why somebody used the function, with a lot of additional code, while it only flags this one AE? So it might lead to additional checks which would not be necessary.

Another point might be that if it is rarely used, we should try to avoid these special cases to stay true to our goal of covering 80% and not every possible case. If a functionality like this is really needed, the users could create their own function and submit it. 

These are my arguments for removing the at_* option, if I understood it correctly :)

R/derive_var_confirmation_flag.R
```diff
@@ -0,0 +1,350 @@
...
+#'
+#'   Whereas, sometimes you might allow for confirmatory observations to occur
+#'   prior to the observation to be flagged. For example, to flag AEs occurring
+#'   on or after seven days before a COVID AE. Thus `join_type = "all"` could be
+#'   used.
+#'
+#'   *Default:* `"after"`
+#'
+#'   *Permitted Values:* `"before"`, `"at_before"`, `"at_after"`, `"after"`,
+#'   `"all"`
```
thomas-neitmann, line 62: I don't think there should be a default value here. I'd rather have people concisely select the `join_type`.

bundfussr: Fine with me. @millerg23 , @rossfarrugia , do you agree?

rossfarrugia: yes agree

tests/testthat/test-derive_var_confirmation_flag.R
bundfussr: I completely agree that the `at_*` option is needed only in rare case and that the function is quite complex.

I added these options
- because they made the `join_type` parameter complete and
- they do not increase the complexity of the function.

I think we do not gain anything if we remove them. But we risk that users need to implement their own function. We can implement these options for free but if users need to create their own function, it will take them a lot of time.

rossfarrugia: I'd like these `at_` options to be removed please Stefan. It may not increase the complexity from the dev perspective but it massively does from a user perspective as it adds a whole new dimension to what is trying to be achieved here by enabling the original record to confirm itself. If ever the user needs this i prefer that they use a simpler solution themselves as Robin suggests. I already feel like this function will rarely be adopted by users given the concept and documentation is so difficult to get your head around. I think the feedback we're most likely to get is to further simplify it rather than extending it.

rossfarrugia (approved):
Updates look good to me and this function will work for our admiralonco and admiralroche needs. I'll be interested to see any user feedback on this one down the line as it is very complex, but then i can't think of any suggestions on how to make simpler.

Issue #1374, opened 2022-08-24
Define the scope of release 0.9.0


duniek:
### Background Information

Create a list of features that would be part of the next release in December.
Highlight which would be the highest priority.

### Definition of Done

List of desired features/ updates available

**Timeline:**  A new release will occur in early December 2022 (release Q4-2022).
**Theme:** Developer and Adoption Focus
**Synopsis:**  

A two prong strategy is adopted.  As admiral sprawls out, we are in need of updating and enhancing our developer documents and processes.  Emphasis will be placed on ensuring that our developer vignettes are up to date with our best practices and that theses are implemented throughout our admiral packages and repos.  At the same time, we also need to ensure that admiral is being adopted at our respective parent companies and the boarder community.  Emphasis will also be placed on updating and creating additional users guides, documentation, video walkthroughs and continuing with our monthly beginner admiral series. 

admiral bugs still receive **the highest priority** to fix and address.  Additional work on new functions, new workflows, etc, with the exception of those identified for this release, should be addressed after we have allocated appropriate resourcing to developer and adoption strategy.  

Issues that address this strategy have been tagged with (release Q4-2022).

rossfarrugia:
@bms63 @thomas-neitmann @millerg23 FYI here's some early ones to add to the list that we'd appreciate consideration to be prioritised for next release from a Roche selfish perspective... 

- https://github.com/pharmaverse/admiral/issues/1338
- https://github.com/pharmaverse/admiral/issues/989 (TRTEMFL) _- i was speaking to Gordon and we were thinking maybe best for @bundfussr to pick this one up as we need to also create an admiralroche version of this later as we have a subtle company-specific addition to the rules here that wouldn't be applicable for admiral_
- https://github.com/pharmaverse/admiral/issues/1337 & https://github.com/pharmaverse/admiral/issues/1336 (@tamarasenior already assigned to)
- https://github.com/pharmaverse/admiral/issues/1041
- using {parttime} as a dependency for the datetime parsing if it's ready in time: https://github.com/pharmaverse/admiral/issues/1315
- bring admiraltemplate in line with admiral latest conventions: https://github.com/pharmaverse/admiraltemplate/issues/37
- all the pieces discussed with @cicdguy around using admiralci directly for the workflows and enabling multi-version docs so that anyone can view the latest site from `devel` anytime without all the links breaking

bms63:
* Creating another demo video for the youtube channel should be added
* Merger Vignette #734 
[...]

PR #1385, opened 2022-08-26
Use Admiral CI workflows

cicdguy:
Use common Admiral CI workflows.
Relates to #1374


cicdguy:
An admin will need to change these settings in the branch protection settings, since the workflow names will change as a result of these common workflows.

Screen Shot 2022-08-26 at 11 21 46 AM

.github/workflows/common.yml
```diff
@@ -0,0 +1,115 @@
...
+  docs:
+    name: Documentation
+    uses: pharmaverse/admiralci/.github/workflows/pkgdown.yml@main
+    if: github.event_name == 'push'
+    with:
+      r-version: $R_VERSION
+      # Whether to skip multiversion docs
+      # Note that if you have multiple versions of docs,
+      # your URL links are likely to break due to path changes
+      skip-multiversion-docs: true
```
cicdguy, line 80: We will set this to `false` following the 0.9.0 release.

.github/workflows/common.yml
```diff
@@ -0,0 +1,115 @@
...
+  # to run this workflow on demand, anytime
+  workflow_dispatch:
+  # 'push' events are triggered when commits
+  # are pushed to one of these branches
+  push:
+    branches:
+      - main
+      - devel
+      - pre-release
+      - test
```
thomas-neitmann, line 18: We don't have a `test` branch.

.github/workflows/common.yml
```diff
@@ -0,0 +1,115 @@
...
+  # 'pull_request' events are triggered when PRs are
+  # created against one of these target branches.
+  pull_request:
+    branches:
+      - main
+      - devel
+      - pre-release
+      - test
```
thomas-neitmann, line 26: See comment above.

.github/workflows/common.yml
```diff
@@ -0,0 +1,115 @@
...
+      # Setting to 'false' will require you to create
+      # an orphan branch called 'badges' in your repository
+      skip-coverage-badges: false
+  templates:
+    name: Check Templates
+    uses: pharmaverse/admiralci/.github/workflows/check-templates.yml@main
+    if: >
+      github.event_name == 'push' || github.event_name == 'pull_request'
+    with:
+      r-version: $R_VERSION
```
thomas-neitmann, line 109: Right now we only run this when a PR is approved as this takes up quite a bit of run time (and thus our free GA minutes).

.github/workflows/common.yml
```diff
@@ -0,0 +1,115 @@
...
+  validation:
+    name: Validation
+    uses: pharmaverse/admiralci/.github/workflows/r-pkg-validation.yml@main
+    if: github.event_name == 'release'
+    with:
+      r-version: $R_VERSION
+  check:
+    name: Check
+    uses: pharmaverse/admiralci/.github/workflows/r-cmd-check.yml@main
+    if: github.event_name == 'pull_request'
```
thomas-neitmann, line 70: I assume that still runs on multiple R versions? What about the package versions? In our current `R CMD check` workflow we use MRAN snapshots from the date of the release of the tested R version. I'd very much like to continue doing that going forward.

thomas-neitmann (changes requested):
Thanks @cicdguy! Please take a look at my comments. In addition please merge in the latest changes from `devel`. That should fix the failing `Check Templates` workflow.

thomas-neitmann (approved):

PR #1481, opened 2022-09-30
Add CRAN Status Monitor

cicdguy:
So this will not go into effect until it's merged into `main`, because `main` is the default branch and scheduled and dispatched workflows are initialized on merge to the default branch.

This workflow can be applied to other Admiral projects, and other CRAN projects in general.


bms63 (approved):
This is so cool.  I love cron

bms63:
Is there anything stopping us from merging this?

thomas-neitmann:
Doesn't seem that way @bms63. You just have to merge it 😉

PR #1544, opened 2022-10-26
Closes #1338 implement feature global option subject keys@devel

### LINKED ISSUE #1338: Feature Request:  Global option program to save maintaining things in many programs
### Feature Idea

Summarizing the below thread of discussions with an agreement from myself, Ben and Thomas over a call:

We intend here to allow for the subject key variables (only) to be controlled by an option that a user could pass once for their study/reporting event etc and it would impact all function calls. Default option would still be `vars(STUDYID, USUBJID)` so we have no breaking change and many users would never need to use this. An example of usage would be say if for a study you needed to use `USUBJID2` instead of `USUBJID`.

The way we plan to achieve this would be to make a new admiral function to set the subject key option using `getOption("admiral_subject_keys", vars(STUDYID, USUBJID))`. By creating a new function for this it would allow us to include a functional reference page with explanatory documentation and examples. 

All of the functions that include the `subject_keys` argument would then use the global option `"admiral_subject_keys"` as default and in the explanation of this argument we could link to the new admiral function documentation replacing the below 2nd sentence:
image

Note: below there were ideas for other areas this could be potentially used, but we wanted to limit it to only places where it saves a substantial amount of user copy and paste across most functions, as by including this we recognise it's another layer of complexity and means developers and QC'ers have to be aware of one additional program.

zdz2101:
Thank you for your Pull Request! We have developed this task checklist from the [Development Process Guide](https://pharmaverse.github.io/admiraldev/main/articles/development_process.html) to help with the final steps of the process. Completing the below tasks helps to ensure our reviewers can maximize their time on your code as well as making sure the admiral codebase remains robust and consistent.   

- [x] Updated relevant unit tests or have written new unit tests - See [Unit Test Guide](https://pharmaverse.github.io/admiraldev/main/articles/unit_test_guidance.html#writing-unit-tests-in-admiral-)
- [x] If you removed/replaced any function and/or function parameters, did you fully follow the [deprecation guidance](https://pharmaverse.github.io/admiraldev/main/articles/programming_strategy.html#deprecation-1)?
- [x] Update to all relevant roxygen headers and examples 
- [x] Update `NEWS.md` if the changes pertain to a user-facing function (i.e. it has an `@export` tag) or documentation aimed at users (rather than developers)


R/admiral_options.R
```diff
@@ -0,0 +1,194 @@
...
+#'
+#' @details
+#' This function allows flexibility for function inputs that may need to be repeated
+#' multiple times in a script, such as for `subject_keys`.
+#'
+#' @author Zelos Zhu
+#'
+#' @return
+#' Currently `subject_keys` returns `vars(STUDYID, USUBJID)` unless user changes the default
+#' using `set_admiral_options()`.
```
bundfussr, line 31: I would phrase it more general to avoid that we have to update it when we add another admiral option. For example, "The value of the specified option"

bundfussr (changes requested):
`period_dataset.R` needs to be updated.

Should we update `vars(STUDYID, USUBJID)` to `get_admiral_options(subject_keys)` in the examples, e.g., in https://pharmaverse.github.io/admiral/reference/derive_vars_merged.html#ref-examples.
@bms63 , @thomas-neitmann , @rossfarrugia , what do you think?

R/admiral_options.R
```diff
@@ -0,0 +1,194 @@
+admiral_default <- new.env(parent = emptyenv())
```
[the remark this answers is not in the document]
zdz2101: I was running into an "cannot change value of locked binding" errors without it, took some inspiration from [R Packages (2e) Section 8.5](https://r-pkgs.org/Data.html#sec-data-state)

R/admiral_options.R
```diff
@@ -0,0 +1,194 @@
...
+#'   missing_value = "N",
+#'   set_values_to = vars(
+#'     PARAMCD = "MDIS",
+#'     PARAM = "Measurable Disease at Baseline"
+#'   ),
+#'   subject_keys = get_admiral_options(subject_keys)
+#' )
+get_admiral_options <- function(input) {
+  # Check for valid input - catch function abuse
+  assert_symbol(enquo(input))
```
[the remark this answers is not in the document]
bundfussr: We have the convention to pass single variables unquoted to a function argument. We did not discuss if it should apply to other objects like options as well. For arguments which expect a keyword like the `mode` argument we expect character scalars.
@bms63 , @thomas-neitmann , what do you think?

R/admiral_options.R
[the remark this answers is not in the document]
bundfussr: I see.

Should we name it more general, e.g., `admiral_env`? Then we can use it for all admiral objects we want to modify.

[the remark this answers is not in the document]
zdz2101: I like this idea, as for other admiral objects we want to modify, do you think it would be a good idea to put this `admiral_env` code-chunk in a "R/aaa.R" script then as well? There may have been some unintended luck involved in naming it "admiral_options.R". In the scenario had I named this script "options_for_admiral.R" instead, scripts that come before alphabetically, wouldn't be able to call `admiral_env` correct?

R/admiral_options.R
```diff
@@ -0,0 +1,194 @@
...
+#'   dataset_add = tu,
+#'   filter_add = TUTESTCD == "TUMIDENT" & VISIT == "SCREENING",
+#'   condition = TUSTRESC == "TARGET",
+#'   false_value = "N",
+#'   missing_value = "N",
+#'   set_values_to = vars(
+#'     PARAMCD = "MDIS",
+#'     PARAM = "Measurable Disease at Baseline"
+#'   ),
+#'   subject_keys = get_admiral_options(subject_keys)
```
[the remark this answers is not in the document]
bundfussr: Yes, for most users these functions work in the background and they won't need to use them and won't need to read the documentation.

I think the documentation is for users who want to change `subject_keys`. In this case I would use `get_admiral_option()` in all places where I need the subject identifiers, e.g., in the example code I stated above. Otherwise, the programs become inconsistent.

R/admiral_options.R
```diff
@@ -0,0 +1,194 @@
...
+#' multiple times in a script, such as for `subject_keys`.
+#'
+#' @author Zelos Zhu
+#'
+#' @return
+#' Modify an admiral option, e.g `subject_keys`, such that it automatically affects downstream
+#' function inputs where `get_admiral_option()` is called such as `derive_param_exist_flag()`.
+#'
+#' @keywords source_specifications
+#' @family source_specifications
```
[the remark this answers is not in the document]
rossfarrugia: i'd vote for under:  "Advanced Functions"/"Admiral Options" - as most users will never need to use this so it kind of is an advanced feature in my mind

R/admiral_options.R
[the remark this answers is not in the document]
bundfussr: Yes, that's correct.

For now I would leave the code-chunk in `admiral_options.R`. If at some point in the future we need to deviate from the alphabetic order, I would rather use the `@include` tag than names like `aaa.R`.

R/admiral_options.R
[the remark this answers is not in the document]
thomas-neitmann: Let's stick to simple `character` here.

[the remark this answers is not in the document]
thomas-neitmann: Really where we want to use unquoted symbols is in cases we are referring to variables inside a dataset, e.g. when using the `new_var` parameter.

R/admiral_options.R
```diff
@@ -0,0 +1,166 @@
...
+    admiral_env$admiral_options$subject_keys <- vars(STUDYID, USUBJID)
+    # admiral_env$admiral_options$future_input <- vars(...) nolint
+
+    ######################################################
+    ### To enhance feature and add inputs as necessary ###
+    ######################################################
+
+    # 1. Add additional options such as future_input as shown commented above
+    # 2. Update @params with future_input in set_admiral_options roxygen documentation
+  }
```
thomas-neitmann, line 164: I would move the setting of default values for options outside this function and directly to where you create the environment. Then you can get rid of the `set_default` parameter.

thomas-neitmann (changes requested):

R/admiral_options.R
```diff
@@ -0,0 +1,166 @@
+admiral_env <- new.env(parent = emptyenv())
+admiral_env$admiral_options <- list()
```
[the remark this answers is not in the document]
zdz2101: I think the intuition for using a list was to be able to adjust quickly for a future use-case where we might have both 
1. another admiral option beyond `subject_keys` 
2. but also happen to use this environment for other admiral objects, as mentioned here by @bundfussr above: 
> 
> Should we name it more general, e.g., `admiral_env`? Then we can use it for all admiral objects we want to modify.

 and the list would be needed to discern and index the admiral options ~line 53, rough sketch below: 
image

All in all, for the single purpose of `subject_keys`, right now I agree it certainly can be simplified.

R/admiral_options.R
zdz2101: Although restarting the R session would restore the defaults anyway, the intuition behind having `set_default` parameter was to allow a quick & easy way for the user to "go back" within the same session. This came up because I thought I needed to clean up after myself because in the roxygen documentation examples I used `set_admiral_options(subject_keys = vars(STUDYID, USUBJID2))` and it could carry downstream to other examples for other functions that call `subject_keys = get_admiral_option(subject_keys)` . This actually came up in the initial drafting process when I was using a `options()/getOption()` based solution instead of this current environment-based one.

R/admiral_options.R
```diff
@@ -0,0 +1,157 @@
...
+
+######################################################
+### To enhance feature and add inputs as necessary ###
+######################################################
+
+# 1. Add additional options such as future_input as shown commented above
+# 2. Update @params with future_input in set_admiral_options roxygen documentation
+# 3. Add future_input into set_admiral_options() formals and body
+
+#' Get the value of an admiral option
```
bundfussr, line 13: Please use title case to be consistent.

R/admiral_options.R
```diff
@@ -0,0 +1,157 @@
...
+  else {
+    default_err_msg <- sprintf(paste(
+      "Invalid function argument, select one of:",
+      enumerate(possible_inputs, quote_fun = dquote, conjunction = "or")
+    ))
+    abort(default_err_msg)
+  }
+}
+
+#' Set the values of admiral options
```
bundfussr, line 78: Please use title case to be consistent.

R/admiral_options.R
```diff
@@ -0,0 +1,157 @@
...
+#'
+#' @details
+#' This function allows flexibility for function inputs that may need to be repeated
+#' multiple times in a script, such as `subject_keys`.
+#'
+#' @author Zelos Zhu
+#'
+#' @return
+#' The value of the specified option unless user changes the default
+#' using `set_admiral_options()`.
```
bundfussr, line 30: I would remove "unless user changes the default using `set_admiral_options()`" because the function returns the value of the option also if it was changed.

tests/testthat/test-admiral_options.R
```diff
@@ -0,0 +1,29 @@
...
+})
+
+## Test 2: common typo gives error ----
+test_that("get_admiral_option Test 2: common typo gives error", {
+  expect_error(get_admiral_option("subject_key"))
+})
+
+## Test 3: unexpected function input for get gives error ----
+test_that("get_admiral_option Test 3: unexpected function input for get gives error", {
+  expect_error(get_admiral_option(subject_keys))
```
bundfussr, line 14: This test does not test the function because `subject_keys` does not exist:
```
Error in assert_character_scalar(option) : 
  object 'subject_keys' not found
```

bundfussr (approved):

rossfarrugia:
@thomas-neitmann this needs an approval from you, so that then @zdz2101 can make the equivalent updates for {admiralonco}: https://github.com/pharmaverse/admiralonco/issues/175

R/admiral_options.R
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  # Return message otherwise, catch typos
+  else {
+    default_err_msg <- sprintf(paste(
+      "Invalid function argument, select one of:",
+      enumerate(possible_inputs, quote_fun = dquote, conjunction = "or")
+    ))
+    abort(default_err_msg)
+  }
```
thomas-neitmann, line 74: The `else` here is redundant. When the `if` is trggered the call to `return()` will exit the function and if not this code will always run even without an `else`.

thomas-neitmann (changes requested):

R/admiral_options.R
thomas-neitmann: Whiel I'm being nit-picky aready as there's no other error message than this one there's no need for the `default` prefix. Just `err_msg` should be fine.

zdz2101: Updated!

R/admiral_options.R
```diff
@@ -0,0 +1,153 @@
...
+
+  if (option %in% possible_inputs) {
+    return(admiral_options[[option]])
+  }
+
+  # Return message otherwise, catch typos
+  err_msg <- sprintf(paste(
+    "Invalid function argument, select one of:",
+    enumerate(possible_inputs, quote_fun = dquote, conjunction = "or")
+  ))
```
thomas-neitmann, line 70: Sorry yet another one... You're not actually using `sprintf()` here. Just `paste()` should suffice.

thomas-neitmann (changes requested):

R/admiral_options.R
zdz2101: Np, updated!

thomas-neitmann (approved):

PR #1628, opened 2022-12-02
[DRAFT] Style only existing files

arkadiuszbeer:


arkadiuszbeer:
It was only for testing

PR #1691, opened 2023-01-27
Fix cran status duplicates

dgrassellyb:
Closes https://github.com/pharmaverse/admiralci/issues/48


bms63:
Ross wants to be taken off these Status alerts.

PR #1753, opened 2023-02-17
Add multi-version pkgdown documentation

cicdguy:
This will create multiversion docs for `admiral`


bms63:
> Note that the real effect will only take place after this has been promoted from `devel` to `main`, as the GHA workflow rules only runs pkgdown documentation on `main`.
> 
> So you won't see the effect after merging this PR.

So I know this kind of violates our lockdown of main, but I'm interested in testing this out before we do a release?  @thomas-neitmann  can we release that restriction to test this out on `main`?  If so, my inner @rossfarrugia is telling me not to this until Monday!

bms63:
Hi @cicdguy and @dgrassellyb.  Are we able to try this out now?  Was hoping to test it out before official release on March 6th.

bms63:
@cicdguy @dgrassellyb Hey all.  I was confused about the slack message.  Is this one ready to go or is the `admiraldev` issue still being investigated.  I actually saw that 0.2 issue in a fork of a admiral repo I did recently.

bms63:
Many thanks @dgrassellyb and @cicdguy for putting this together.  I think a lot of Pharma companies will like this versioning of the website as some drug studies might be locked into a particular version of admiral for several years.  While the pkg documentation is within the package it isn't always so pleasant to view and I Find the website to be very helpful and clean to look at.

PR #2026, opened 2023-07-20
Update templates.yml

bms63:


bms63:
Hi @cicdguy and @galachad, it was decided to no longer have these templates fire when approved but when a PR is created/updated instead of doing the snapshot tests.

bms63:
I had to look up what DX meant.  

image

Issue #2003, opened 2023-07-07
PROPOSAL: Simplify git branching structure


ddsjoberg:
### Background Information

The R community has a typical branching structure used among the vast majority of open source package projects.  I propose we update to follow the community structure. The community uses the main branch for the latest dev version of the package. I think we recently saw this when Hadley tried to make an update to the repo, but couldn't find the correct place to make the change.

### Definition of Done

- [ ] make main the default branch
- [ ] remove the devel branch
- [ ] clean up stale branches

bms63:
im just wondering what would be impacted by dropping the `devel` branches from `admiral`, `admiraldev` and `admiral.test/admiraldata`

I think pretty minimal??

* Impact to staged.dependencies process
* Guidance on how to patch a bug and re-release
* Website deployment with github release/tag
* others?

bundfussr:
> https://bookdown.netlify.app/hadley/r-pkgs/release.html (also https://r-pkgs.org/release.html for a cleaner interface) is what the R community uses. Semver 2.0 is the gold standard, however.

I couldn't find anything regarding branches.

How would you do a hotfix? Create a hotfix branch from the last release tag, do the fixes there, submit from the hotfix branch, update the website from the hotfix branch, do a release from the hotfix branch, and merge it back to main?

bundfussr:
But if we create a new release from main, it may contain changes we do not want to release. For example, assume we have released 0.12.0, then we add some changes for 0.13.0 to main, and then we need a hotfix for 0.12.0. Version 0.12.1 should contain the updates for the hotfix but not the updates for 0.13.0.

And how would staged dependencies determine the correct version of the dependencies? For example, the workflows running for the hotfix should use the latest released version of the dependency and not main.

ddsjoberg:
That's a good point @cicdguy !

Regarding hotfixes in general: If we are ever in need of an immediate patch, the vast majority of the time we can make the fix and also release it along with any new features/fixes in main without any need for a hotfix branch

bundfussr:
> Regarding hotfixes in general: If we are ever in need of an immediate patch, the vast majority of the time we can make the fix and also release it along with any new features/fixes in main without any need for a hotfix branch

[...]

PR #2116, opened 2023-09-20
Using tags for hotfixes

bms63:
I created branch from tag

image

* called it `v0_12_1hofix` branch
* doing first bug fix with `v0_12_1_bugfix1`
* could do another bug fix with `v0_12_1_bugfix2`
* etc...

Assuming that I can then can do a release via my `v_12_1_hofix` 

image

Once this is deployed we to CRAN - new website.  We can merge `v_12_1_hofix` into `main`.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,3 +1,4 @@
+# This wizard shit is hard
 # Set renv profile base on R version.
 .get_dependencies <- function(project_dir) {
 


bms63:
> Language?

haha - yeah won't be making it into admiral :)

zdz2101:
Btw @bms63 if you run into issues with deleting tags later https://stackoverflow.com/questions/5480258/how-can-i-delete-a-remote-tag

bms63:
This all seems pretty straightforward to me and in the end similar to how we did it with `pre-release` and `hotfix` branches.

Create Branch off tag at version of `admiral` we want to patch.

Continue to Branch off newly created patch branch for different hotfixes, doc updates, etc

merge all back into patch branch 

create release from patch branch 

merge patch branch back into main

...thinking the patch branch has to remain with us forever (can set a rule not to delete them)

bms63:
I guess if someone wanted to patch `1.0`  while `2.0` is out we could do that with this new strategy.  So maintain two version of admiral if there is some crazy world we live in that requires that hellscape.

manciniedoardo:
LGTM - let's make sure to update [this section](https://pharmaverse.github.io/admiraldev/devel/articles/release_strategy.html#hot-fix-release) of our documentation once this process is set in stone.

manciniedoardo (approved):

bundfussr:
Why do you want to include the version in the name of the patch branch?
Why not naming it `patch`?
Then we have _only one_ additional branch which remains with us forever. Otherwise, we have one additional branch for each hotfix.

And do we need to add the version to the name of the feature branches? I would just use our normal feature branch naming convention, i.e., instead of `v0_12_1_bugfix2` use `3456_bugfix2` where `3456` is the issue number.

bundfussr (changes requested):
Could we simplify it? (see my comment above)

ddsjoberg:
> LGTM - let's make sure to update [this section](https://pharmaverse.github.io/admiraldev/devel/articles/release_strategy.html#hot-fix-release) of our documentation once this process is set in stone.

FYI @manciniedoardo I have an open PR with those updates already 🕺🏼

bundfussr:
> > LGTM - let's make sure to update [this section](https://pharmaverse.github.io/admiraldev/devel/articles/release_strategy.html#hot-fix-release) of our documentation once this process is set in stone.
> 
> FYI @manciniedoardo I have an open PR with those updates already 🕺🏼

I would prefer using a single `patch` branch as suggested in the PR.

@bms63 , @manciniedoardo , what do you think?

zdz2101:
> > > LGTM - let's make sure to update [this section](https://pharmaverse.github.io/admiraldev/devel/articles/release_strategy.html#hot-fix-release) of our documentation once this process is set in stone.
> > 
> > 
> > FYI @manciniedoardo I have an open PR with those updates already 🕺🏼
> 
> I would prefer using a single `patch` branch as suggested in the PR.
> 
> @bms63 , @manciniedoardo , what do you think?

Might be easier to use `patch` notation to trigger all the CI/CD workflows, since as of right now we would need to add it back into all the github actions on `admiralci` to trigger workflows

bundfussr:
> Might be easier to use `patch` notation to trigger all the CI/CD workflows, since as of right now we would need to add it back into all the github actions on `admiralci` to trigger workflows

Good point.

By the way, how do the workflows know which version of the upstream dependencies they should use? In the past it was `main`. With the new branching strategy it should be the latest released version. But the tags may differ depending on the dependency. For example, it could be `v0.1.0` for pharmaversesdtm and `v0.5.0` for admiraldev. I think staged.dependecies can not handle this because using a branch name like `1234_bugfix@v0.1.0@v0.5.0` would be ambiguous.

@cicdguy , @ddsjoberg , any ideas?

bms63:
Hey all,

So rough instructions for a hotfix"

- [ ] Branch off latest tagged release
- [ ] Call the branch `patch`
- [ ] Bug Fixes are applied to the `patch` branch via `issue-number-patch-fix2`, `issue-number-patch-fix2`, etc and merged into `patch`
- [ ] Submit `patch` to CRAN to make sure it is accepted
- [ ] Complete release via `patch` branch to redeploy website
- [ ] Merge `patch` branch into `main`
- [ ] Delete `patch` branch
- [ ] ...
- [ ] Profit

Other items to attend to:

- [ ] CI workflows need patch trigger re-introduced
- [ ] Release workflow needs patch branch trigger
- [ ] Update documentation on doing a hotfix release

bundfussr:
We should not delete the `patch` branch because we will release from it. (See Dinakars comment: https://github.com/pharmaverse/admiral/issues/2003#issuecomment-1630984011.)

It should be protected and we need to decide when we update it. We could update it after each release. Then it is always ready for a hotfix. Or we update it when we start a hotfix.

bms63:
> We should not delete the `patch` branch because we will release from it. (See Dinakars comment: [#2003 (comment)](https://github.com/pharmaverse/admiral/issues/2003#issuecomment-1630984011).)
> 
> It should be protected and we need to decide when we update it. We could update it after each release. Then it is always ready for a hotfix. Or we update it when we start a hotfix.

Okay so `patch` branch hangs around.  I updated the punch list.  I think we are in good shape now to proceed.

bms63:
I made an issue to update programming docs on how to make a hotfix under our new GitHub Flow Strategy.  Going to clsoe this PR

PR #2148, opened 2023-10-03
Update common workflow

cicdguy:
Because we're now using a single-branch GitHub Flow strategy, we don't need other branches that we need to act on.

We also don't want to build pages for `main`.


ddsjoberg:
@cicdguy we may still want to build the pkgdown site on main. There is a pkgdown option to build the dev site and save it in a subfolder named `dev`. I think this newly option is turned on in one of the open PRs.

Is the primary site (latest release) now being hosted from the root of the `gh-pages` branch?

.github/workflows/common.yml
```diff
@@ -15,17 +15,11 @@ on:
       - "v*"
     branches:
       - main
-      - devel
-      - patch
```
bms63, line 19: we actually need the patch one as we still don't want to release everything from `main` when doing a patch/bug fix

ddsjoberg: After you release from patch, you'd merge to main and the site would be built then. So do we need the patch pkgdown builds?

bms63: but i think we still want the workflows to trigger when PRing into `patch`?  IF we remove `patch` then it won't trigger any of the checks - I thought??

bms63 (approved):
Let's see what happens!

bundfussr:
@cicdguy , @bms63 , do we have a separate PR to run the checks in `patch` (and for releases) with the latest released version of the upstream dependencies (instead of the version from `main`)? Or is this already solved?

bms63:
The checks are running when we PR into `patch`.  This update/PR would of removed them, but patch was added back in.

We no longer make the `main` version of the website - which might impact the `dev` version being created.  Not completely sure!!

PR #2206, opened 2023-10-30
Use a newer version of R for pkgdown

cicdguy:


ddsjoberg (approved):
✅

bms63:
Should we just upgrade all workflows to `4.3`

PR #2213, opened 2023-11-01
Add secret for `gert`'s date fetcher

cicdguy:
`gert` is used for the articles that has the package version links. It needs a GitHub PAT to get release dates from the GitHub releases API.


ddsjoberg (approved):
🎆 looks great! 🤞🏼

PR #2261, opened 2023-11-29
Closes #2255 add uppercase spellings of in_unit/out_unit

### LINKED ISSUE #2255: General Issue: Time Units Set by Duration Functions
### Background Information

The following issue was reported by a user:

> Whilst creating `ADCM.LDRELTMU` I had an error with metacore that `"MINUTES"` is not in the controlled terminology. The actual value is `"min"`. Updating the derivation as below is only partly successful, as the unit also needs to be lowercased.
> ```
>   derive_vars_duration(
>    new_var = LDRELTM,
>    new_var_unit = LDRELTMU,
>    start_date = ASTDTM,
>    end_date = LDOSEDTM,
>    in_unit = "seconds",
>    out_unit = "min",
>    add_one = FALSE
>  )
>```

The following CDISC codelist was used
image

Currently, the unit variable is set to the upper case value of `out_unit`.

Possible solutions:
- Set the unit variable to the (unchanged) value of `out_unit` and accept a variety of values like `"MIN"`, `"min"`, `"MINUTES"`, `"minutes"`, `"MINUTE"`, `"minute"`, ... (Should the same values be valid for `in_unit` then?)
- Accept `"years"`, `"months"`, `"days"`, `"hours"`, `"minutes"`, `"seconds"` for `out_unit` only and add an argument for the value of the unit variable.
- Accept `"years"`, `"months"`, `"days"`, `"hours"`, `"minutes"`, `"seconds"` for `out_unit` only and add an argument expecting a function which maps the `out_unit` values to other values like the codelist values. (default value?)

@bms63 , @manciniedoardo , what do you think?

### Definition of Done

_No response_

zdz2101:
Thank you for your Pull Request! We have developed this task checklist from the [Development Process Guide](https://pharmaverse.github.io/admiraldev/articles/development_process.html) to help with the final steps of the process. Completing the below tasks helps to ensure our reviewers can maximize their time on your code as well as making sure the admiral codebase remains robust and consistent.   

- [x] Updated relevant unit tests or have written new unit tests, which should consider realistic data scenarios and edge cases, e.g. empty datasets, errors, boundary cases etc. - See [Unit Test Guide](https://pharmaverse.github.io/admiraldev/articles/unit_test_guidance.html#tests-should-be-robust-to-cover-realistic-data-scenarios)
- [x] If you removed/replaced any function and/or function parameters, did you fully follow the [deprecation guidance](https://pharmaverse.github.io/admiraldev/articles/programming_strategy.html#deprecation)?
- [x] Review the [Cheat Sheet](https://github.com/pharmaverse/admiral/blob/main/inst/cheatsheet/admiral_cheatsheet.pdf). Make any required updates to it by editing the file `inst/cheatsheet/admiral_cheatsheet.pptx` and re-upload a PDF version of it to the same folder.
- [x] Update to all relevant roxygen headers and examples, including keywords and families. Refer to the [categorization of functions](https://pharmaverse.github.io/admiraldev/articles/programming_strategy.html#categorization-of-functions) to tag appropriate keyword/family.
- [x] Update `NEWS.md` under the header `# admiral (development version)` if the changes pertain to a user-facing function (i.e. it has an `@export` tag) or documentation aimed at users (rather than developers)
- [x] Build admiral site `pkgdown::build_site()` and check that all affected examples are displayed correctly and that all new functions occur on the "[Reference](https://pharmaverse.github.io/admiral/cran-release/reference/index.html)" page. 


bms63 (commented):
I think we also needed some other weird variations - like `s` for seconds.

image

R/compute_duration.R
```diff
@@ -154,16 +154,23 @@ compute_duration <- function(start_date,
   # Checks
   assert_date_vector(start_date)
   assert_date_vector(end_date)
-  assert_character_scalar(in_unit, values = valid_time_units())
+  assert_character_scalar(in_unit, values = c(
```
bundfussr, line 157: `in_unit` should accept the same values as `out_unit` except for values referring to weeks.

R/compute_duration.R
```diff
@@ -154,16 +154,23 @@ compute_duration <- function(start_date,
...
+    valid_time_units(),
+    toupper(valid_time_units())
+  ))
   assert_character_scalar(type, values = c("interval", "duration"))
   assert_character_scalar(out_unit, values = c(
-    valid_time_units(), "weeks",
-    "min", "sec"
+    valid_time_units(),
+    toupper(valid_time_units()),
+    "weeks", "min", "sec"
```
bundfussr, line 165: What about values like `"year"`, `"yrs"`, `"wk"`, `"month"`, ...?

Maybe option 1 was not a good decision. This could become a source of frequent updates.

bundfussr (changes requested):
- `toupper()` should be removed from the `mutate()` call in line 243 of `derive_vars_duration.R`.
- The `age_unit` argument in `derive_vars_aage()` needs to be updated.

R/compute_duration.R
bms63: I thought we were only updating to the ones allowed in the CDSIC image?

manciniedoardo: Yeah maybe we stick to those just to cover our bases with regards to this codelist

bundfussr: I think it is not mandatory to use the codelist and the codelist is extensible. So users may want to use different values for the units and ask us to update the function. I wonder if we should update the function such that any value can be used.

bms63: Perhaps we have a soft check/message on the values that are shown in the codelist with an option to allow an extensible codelist.  

So a user would be prompted with a message that says "These are the common values used in this function.  If you wish to use additional values please mark `extensible = TRUE`  to suppress this message

bundfussr: I think this will not work because we use the `out_unit` value not only for setting the unit variable but also for calling lubridate functions. So we need to translate these values before passing them to the lubridate function.

zdz2101: I think where it gets a little tricky is we rely on passing whatever we deem as acceptable unit values into the `unit` argument inside the `lubridate` package, which after investigating a bit, they have some pretty "loose" acceptable terms (they generally accept the singular or plural versions of the whole word of the first three letters, in particular see second screenshot for clear demo)

image

image

My proposal if we want to assist in mapping to the proper lubridate functions we limit our scope to:

- the standard ones (case-agnostic) singular & plural forms so week, WEEK, etc. 
- Y, W, D, H, S (can't do M because month vs minutes) 
- yr(s), mo(s), wk(s), hr(s)

R/compute_duration.R
```diff
@@ -151,19 +151,35 @@ compute_duration <- function(start_date,
...
+    years = c("year", "years", "yr", "yrs", "y"),
+    months = c("month", "months", "mo", "mos"),
+    weeks = c("week", "weeks", "wk", "wks", "w"),
+    days = c("day", "days", "d"),
+    hours = c("hour", "hours", "hr", "hrs", "h"),
+    minutes = c("minute", "minutes", "min", "mins"),
+    seconds = c("second", "seconds", "sec", "secs", "s")
+  )
+  in_unit <- names(mapping[mapping %in% tolower(in_unit)]) %>% str_replace(., '[0-9]', '')
+  out_unit <- names(mapping[mapping %in% tolower(out_unit)]) %>% str_replace(., '[0-9]', '')
```
zdz2101, line 164: @bms63 @bundfussr  @manciniedoardo Take a look here at lines 154-164, which make it easily modifiable what we accept/want to map onto our duration calculations, see corresponding tests in `test-compute_duration.R`

zdz2101: Ahh yes totally forgot that was introduced, we would have to upversion of description files for dplyr (>=1.1.0) for that though

bundfussr: > Ahh yes totally forgot that was introduced, we would have to upversion of description files for dplyr (>=1.1.0) for that though

If we do not want to upversion, we could use `case_match()`. @bms63 , @manciniedoardo , what do you think?

bms63: I'm hesitant to fix to 1.1.0 as it has only been out a year.  Maybe simpler isn't the solution here.

zdz2101: Well both solutions are already drafted, only a matter of deciding which to proceed with, for easier review purposes I've linked the commit associated with both below:

1. no `case_match` from [this commit](https://github.com/pharmaverse/admiral/pull/2261/commits/d6c550eee01665b9430260381cdadf1639b0497c)

2. with `case_match`  where it stands [right now](https://github.com/pharmaverse/admiral/pull/2261/files)

bms63: Given the nature of Pharma I would be want to pick option 1.

bundfussr: > If we do not want to upversion, we could use `case_match()`.

Sorry, I meant `case_when()`. I thinks even with `case_when()` the code is much easier to read.

Furthermore, as we need this mapping more than once I would create a (non-exported) function for it.

R/derive_vars_aage.R
```diff
@@ -93,6 +93,9 @@ derive_vars_aage <- function(dataset,
   start_date <- assert_symbol(enexpr(start_date))
   end_date <- assert_symbol(enexpr(end_date))
   assert_data_frame(dataset, required_vars = expr_c(start_date, end_date))
+
+  age_unit <- valid_time_mappings(age_unit)
```
bundfussr, line 97: This could create confusing error messages. For example
```
> derive_vars_aage(data, age_unit = "wks")
# A tibble: 1 × 4
  BRTHDT     RANDDT      AAGE AAGEU
  <date>     <date>     <dbl> <chr>
1 1984-09-06 2020-02-24  1850 weeks
```
works. But with a typo:
```
> derive_vars_aage(data, age_unit = "wls")
Error in `assert_character_scalar()` at admiral/R/derive_vars_aage.R:99:2:
! `age_unit` must be one of 'years', 'months', 'weeks', 'days', 'hours', 'minutes' or 'seconds' but is 'wls'
```
the error message suggests that `"wks"` is not valid.

Furthermore, the unit variable should be set to the specified value. I.e., in the first example `AAGEU` should be `"wks"` instead of `"weeks"`.

R/compute_duration.R
```diff
@@ -212,3 +219,58 @@ compute_duration <- function(start_date,
...
+#'
+#' @param time_unit input unit
+#'
+#' @return standardized term if mapping is available
+#'
+#' @keywords internal
+#' @family internal
+#'
+#' @noRd
+valid_time_mappings <- function(time_unit) {
```
bundfussr, line 233: According to our programming strategy function names should start with a verb. So it should be something like `get_unified_time_unit()`.

R/compute_duration.R
```diff
@@ -37,7 +37,7 @@
 #'   Default: 'days'
 #'
 #'   Permitted Values: 'years', 'months', 'weeks', 'days', 'hours', 'minutes',
-#'   'min', 'seconds', 'sec'
+#'   'min', 'seconds', 'sec', capitalized versions of these terms is accepted too
```
bundfussr, line 40: This should be updated as we accept even more values now. (`in_unit` and `age_unit` should be updated as well)

bundfussr (changes requested):

DESCRIPTION
```diff
@@ -40,7 +40,7 @@ Depends:
     R (>= 4.0)
 Imports:
     admiraldev (>= 0.5.0.9000),
-    dplyr (>= 0.8.4),
+    dplyr (>= 1.0.5),
```
bms63, line 43: ummm...so this update removes capability with dplyr less than 1.0.5?  Maybe we should say that in our news now??

Any chance we could still work with that version or "are the times they is changing"??

zdz2101: I actually just grabbed it from the `DESCRIPTION` file of `admiraldev` but this has technically been true for a while because we use functions like `across()` which was introduced at `1.0.0` and gradually improved

bms63: well then... i guess we just continue on with this update

R/derive_vars_aage.R
```diff
@@ -93,10 +100,18 @@ derive_vars_aage <- function(dataset,
   start_date <- assert_symbol(enexpr(start_date))
   end_date <- assert_symbol(enexpr(end_date))
   assert_data_frame(dataset, required_vars = expr_c(start_date, end_date))
-  assert_character_scalar(
-    age_unit,
-    values = c("years", "months", "weeks", "days", "hours", "minutes", "seconds")
-  )
+
+  age_unit <- get_unified_time_unit(age_unit)
```
bundfussr, line 104: This should be removed (and `case_sensitive = FALSE` be used in the `assert_character_scalar()` call). Otherwise the unit variable is not set correctly. Consider for example
```
> derive_vars_aage(data, age_unit = "wks")
# A tibble: 1 × 4
  BRTHDT     RANDDT      AAGE AAGEU
  <date>     <date>     <dbl> <chr>
1 1984-09-06 2020-02-24  1850 weeks
```
`AAGEU` should be `"wks"`.

bundfussr (changes requested):

bundfussr:
@zdz2101 , @bms63 , @manciniedoardo , should we change the default of `out_unit` to `DAYS` (in `derive_vars_duration()`) and the default of `age_unit` to `YEARS` (in `derive_vars_aage()`)?
Then we would avoid breaking backward compatibility.

PR #2298, opened 2023-12-19
demo of using tags to patch/hotfix for a release

bms63:
@bundfussr this is demoing creating a patch branch off a tag.  This way we don't need to maintain a patch branch going forward. 

I just followed this - https://stackoverflow.com/questions/10940981/how-to-create-a-new-branch-from-a-tag

I think we did this recently with some of the `0.12.x` patching.  @zdz2101 do you recall if we branched off a tag?

It seems to work!

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,3 +1,7 @@
+# admiral 1.0.1
+
+* Patching something
+
 # admiral 1.0.0
 
 ## New Features


bms63:
So I had to make two branches off the `v1.0.0` tag so we could track the changes and it ins't getting merged to `main`

ONce this is completed we can release from this demo branch and merge the change into `main`

bms63:
code for creating the branches 
image

bundfussr:
As far as I remember the issue was not creating a branch from a tag. I think @cicdguy mentioned that using the patched version issues warnings if the branch was deleted.

By the way, why do you want to get rid of the `patch` branch?

bms63:
> As far as I remember the issue was not creating a branch from a tag. I think @cicdguy mentioned that using the patched version issues warnings if the branch was deleted.
> 
> By the way, why do you want to get rid of the `patch` branch?

I'm just focused on trying to simplify things, but if the solution to "our current solution" is complicated then we can definitely keep the patch branch.

Issue #2403, opened 2024-04-10
Check for other calls to `rlang::abort()` and update with cli


ddsjoberg:

ddsjoberg:
HI @bundfussr , we can close this issue now, right?

bundfussr:
> HI @bundfussr , we can close this issue now, right?

Yes, all calls should be covered in https://github.com/pharmaverse/admiral/issues/2413.

PR #2412, opened 2024-04-17
Related #2403: cli error message updates Part 1

ddsjoberg:
There are _a lot_ of calls to `rlang::abort()` that need updating. Best to tackle this in parts.

----

Thank you for your Pull Request! We have developed this task checklist from the [Development Process Guide](https://pharmaverse.github.io/admiral/CONTRIBUTING.html#detailed-development-process) to help with the final steps of the process. Completing the below tasks helps to ensure our reviewers can maximize their time on your code as well as making sure the admiral codebase remains robust and consistent.   

Please check off each taskbox as an acknowledgment that you completed the task or check off that it is not relevant to your Pull Request. This checklist is part of the Github Action workflows and the Pull Request will not be merged into the `main` branch until you have checked off each task.

- [x] Place Related #2403 into the beginning of your Pull Request Title (Use Edit button in top-right if you need to update)
- [x] Updated relevant unit tests or have written new unit tests, which should consider realistic data scenarios and edge cases, e.g. empty datasets, errors, boundary cases etc. - See [Unit Test Guide](https://pharmaverse.github.io/admiraldev/articles/unit_test_guidance.html#tests-should-be-robust-to-cover-realistic-data-scenarios)
- [x] If you removed/replaced any function and/or function parameters, did you fully follow the [deprecation guidance](https://pharmaverse.github.io/admiraldev/articles/programming_strategy.html#deprecation)?
- [x] Review the [Cheat Sheet](https://github.com/pharmaverse/admiral/blob/main/inst/cheatsheet/admiral_cheatsheet.pdf). Make any required updates to it by editing the file `inst/cheatsheet/admiral_cheatsheet.pptx` and re-upload a PDF version of it to the same folder.
- [x] Update to all relevant roxygen headers and examples, including keywords and families. Refer to the [categorization of functions](https://pharmaverse.github.io/admiraldev/articles/programming_strategy.html#categorization-of-functions) to tag appropriate keyword/family.
- [x] Update `NEWS.md` under the header `# admiral (development version)` if the changes pertain to a user-facing function (i.e. it has an `@export` tag) or documentation aimed at users (rather than developers)


R/compute_scale.R
```diff
@@ -77,19 +77,21 @@ compute_scale <- function(source,
...
   if (!is.null(target_range) && is.null(source_range)) {
-    abort(paste0(
-      "argument `source_range` is missing, with no default, but ",
-      "`target_range` is not missing\nEither both or neither ",
-      "argument should exist"
-    ))
+    cli_abort(
+      c("Argument {.arg source_range} is missing with no default
+         and {.arg target_range} is not missing.",
+        "i" = "Either both or neither argument should by specified."
```
millerg23, line 83:
```suggestion
        "i" = "Either both or neither argument should be specified."
```

R/compute_scale.R
```diff
@@ -77,19 +77,21 @@ compute_scale <- function(source,
...
   if (!is.null(source_range) && is.null(target_range)) {
-    abort(paste0(
-      "argument `target_range` is missing, with no default, but ",
-      "`source_range` is not missing\nEither both or neither ",
-      "argument should exist"
-    ))
+    cli_abort(
+      c("Argument {.arg target_range} is missing with no default
+         and {.arg source_range} is not missing.",
+        "i" = "Either both or neither argument should by specified."
```
millerg23, line 92:
```suggestion
        "i" = "Either both or neither argument should be specified."
```

R/compute_scale.R
ddsjoberg: Thanks for the catch! Updated!

millerg23:
> Should the `class` argument be specified in the `cli_abort()` calls?

How would we name the classes, could use function name, but what if there is more than one check in the function. Would we ever use the class? I can see benefit of having class in assertion functions, so we can just check if check is triggered without checking message. But not sure we would use the class?

bundfussr:
> > Should the `class` argument be specified in the `cli_abort()` calls?
> 
> How would we name the classes, could use function name, but what if there is more than one check in the function. Would we ever use the class? I can see benefit of having class in assertion functions, so we can just check if check is triggered without checking message. But not sure we would use the class?

Good questions!
The use case I had in mind was where you have a company or study specific function which calls an admiral function and you want to catch the error to provide a more user-friendly error message. However, with a class you could easily catch the error but for the message you still need to parse the original message to get out what went wrong and generate the new message. We have this problem even with our `assert_*()` functions. In `assert_terms()` I had to call `assert_data_frame()` twice (once for checking it's a data frame and again for checking if the required variables are present) to be able to update the error message.
Maybe it does not help much to set the class with our current approach.

zdz2101 (commented):
I think adding the class would probably be more helpful for our testing suite/for developers more so than users, personally I think with regard to this particular issue/topic with regard to just the error messaging aspect, this looks good to me, we can add a separate issue/have a discussion if we think adding the class is appropriate later on?

R/call_derivation.R
```diff
@@ -229,18 +229,15 @@ call_derivation <- function(dataset = NULL, derivation, variable_params, ...) {
...
   }
   duplicate_params <- get_duplicates(names(args))
   if (length(duplicate_params) >= 1L) {
-    err_msg <- sprintf(
-      "The following parameters have been specified more than once: %s",
-      enumerate(duplicate_params)
-    )
-    abort(err_msg)
+    cli_abort("The following parameters have been specified more than once:
+               {.val {duplicate_params}}.")
```
kaz462, line 240:
```suggestion
    cli_abort("The following parameter{?s} ha{?s/ve} been specified more than once:
               {.val {duplicate_params}}.")
```

R/compute_age_years.R
```diff
@@ -52,11 +52,13 @@ compute_age_years <- function(age,
...
-      "`age`, but there are ", length(age), " values in `age` and ",
-      length(age_unit), " values in `age_unit`."
-    ))
+    cli_abort(
+      c("Argument {.arg age_unit} must be a single string or a vector of the
+       same length as {.arg age}",
+        "i" = "There are {.val {length(age)}} values in {.arg age} and
+              {.val {length(age_unit)}} values in {.arg age_unit}."
+      )
+    )
```
kaz462, line 61:
```suggestion
    cli_abort(
      c("Argument {.arg age_unit} must be a single string or a vector of the
       same length as {.arg age}",
        "i" = "There are {.val {length(age)}} value{?s} in {.arg age} and
              {.val {length(age_unit)}} value{?s} in {.arg age_unit}."
      )
    )
```

bundfussr:
I have added the changes suggested by @kaz462 to https://github.com/pharmaverse/admiral/pull/2417.

Should we cancel this PR as the changes are included also in https://github.com/pharmaverse/admiral/pull/2417?

bms63:
> I have added the changes suggested by @kaz462 to #2417.
> 
> Should we cancel this PR as the changes are included also in #2417?

Yes lets cancel this one - it is getting a little messy.

bms63:
@bundfussr do you mind closing as you are close to this process

PR #2489, opened 2024-08-06
Dependency conformance tests

cicdguy:
This adds an "on-demand" aka manually triggerable workflow that will allow you to test whether the package conforms with various dependency sets. This workflow uses the [`verdepcheck` GitHub Action](https://github.com/insightsengineering/r-verdepcheck-action).

### DIFF
--- a/.github/workflows/verdepcheck.yml
+++ b/.github/workflows/verdepcheck.yml
@@ -1,0 +1,23 @@
+---
+name: Dependency Conformance Test
+
+on:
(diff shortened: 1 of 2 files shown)


manciniedoardo:
FYI @bundfussr @bms63

bms63 (commented):
Will this only trigger when we do [skip vbump]?

bms63:
Hey @cicdguy will this trigger only with [skip vbump]?

PR #2932, opened 2025-12-19
Test joins removed

bms63:
@bundfussr unsure if the remotes is working - I think the CI action is getting admiraldev from main and not from the declared remotes branch.

my local build with the admiraldev branch is working nicely

image

### DIFF
--- a/staged_dependencies.yaml
+++ b/staged_dependencies.yaml
@@ -2,8 +2,6 @@
 upstream_repos:
 - repo: pharmaverse/pharmaversesdtm
   host: https://github.com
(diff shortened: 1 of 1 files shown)


bundfussr:
You need to remove `admiraldev` from `staged_dependencies.yaml`.

bms63:
It looks like we are good with the removal of joins from admiraldev.  I can just close this PR and we can merge in https://github.com/pharmaverse/admiraldev/pull/531

bms63:
FYI @manciniedoardo removal of joins tested in admiral - looks to be in good shape

